PR #24, opened 2015-01-09
added delay for refresh campaigns and resources to after events in queue...

dominicmarmionswrve:
... are sent

https://swrvedev.jira.com/browse/SWRVE-6888

@Sergio-Mira Can you take a look please?

### DIFF
--- a/SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveImp.java
+++ b/SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveImp.java
@@ -1151,7 +1151,13 @@ abstract class SwrveImp<T, C extends SwrveConfigBase> {
             final SwrveBase<T, C> swrve = (SwrveBase<T, C>) this;
             swrve.sendQueuedEvents();
             eventsWereSent = false;
-            swrve.refreshCampaignsAndResources();
+            ScheduledExecutorService timedService = Executors.newSingleThreadScheduledExecutor();
+            timedService.schedule(new Runnable() {
+                @Override
+                public void run() {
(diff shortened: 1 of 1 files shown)


SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -1182,7 +1189,7 @@ protected void startCampaignsAndResourcesTimer(boolean sessionStart) {
             public void run() {
                 checkForCampaignAndResourcesUpdates();
             }
-        }, campaignsAndResourcesFlushRefreshDelay.longValue(), campaignsAndResourcesFlushFrequency.longValue(), TimeUnit.MILLISECONDS);
+        }, 0l, campaignsAndResourcesFlushFrequency.longValue(), TimeUnit.MILLISECONDS);
```
Sergio-Mira: Why has this setting changed too? I thought the only problem was that it was not waiting (the code above)

dominicmarmionswrve: Because we want to start the campaignsAndResourcesExecutor straight away. There's no need for it to add a delay here. Let the delay happen directly after queued events (and refresh resources) have been sent inside of checkForCampaignAndResourcesUpdates method. 

Essentially this whole pull request is about movign the delay from starting the campaignsAndResourcesExecutor to inside the checkForCampaignAndResourcesUpdates.

Sergio-Mira: Cool

Sergio-Mira:
:+1: :rocket:

PR #26, opened 2015-03-03
7802 Replace JARs with Gradle dependencies

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-7802


dominicmarmionswrve:
:+1:

dominicmarmionswrve:
@Sergio-Mira Sorry, I know I gave :+1: , but I think the following line needs to be removed from SwrveSDK/build.gradle and SwrveGoogleSDK/build.gradle:
	compile 'com.android.support:support-v4:21.0.0'
because it should already be in SwrveCommonSDK/build.gradle

PR #31, opened 2015-03-12
7727 remove init and use activity context

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-7727

private changes:
https://github.com/Swrve/swrve-sdk/pull/791


dominicmarmionswrve:
@Sergio-Mira can you take a look please?

SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -1093,6 +1089,7 @@ public void purchase(String item, String currency, int cost, int quantity) {
         }
     }
 
+    @Override
```
Sergio-Mira: ~ ~ Why the overrides?

Sergio-Mira:
:+1:

SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveBase.java
dominicmarmionswrve: If the interface method signature changes then this makes it far easier to refactor.

PR #33, opened 2015-03-18
Fix for dialog between activities

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-8073

### DIFF
--- a/SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveImp.java
+++ b/SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveImp.java
@@ -917,13 +917,13 @@ abstract class SwrveImp<T, C extends SwrveConfigBase> {
         }
     }
 
-    protected void unbindAndShutdown() {
+    protected void unbindAndShutdown(Activity ctx) {
         // Reduce the references to the SDK
         int counter = bindCounter.decrementAndGet();
         // Remove the binding to the current activity, if any
         this.activityContext = null;
 
-        removeCurrentDialog();
+        removeCurrentDialog(ctx);
 
         // Check if there are no more references to this object
         if (counter == 0) {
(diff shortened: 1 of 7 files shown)


Sergio-Mira:
@dominicmarmionswrve @0vertone Could you have a look?

SwrveCommonSDK/src/main/java/com/swrve/sdk/messaging/view/SwrveDialog.java
```diff
@@ -71,4 +74,11 @@ private void goneAway() {
...
         }
     }
+
+    public Activity getParentActivity() {
+        if (parentActivity != null) {
+            return parentActivity.get();
+        }
+        return null;
+    }
 }
```
dominicmarmionswrve: Does the following do this anyway? http://developer.android.com/reference/android/app/Dialog.html#getOwnerActivity()

Sergio-Mira: It doesn't work properly with custom dialogs

dominicmarmionswrve: If we call setOwnerActivity(activity) in the constructor of our custom SwrveDialog, then getOwnerActivity should work.

SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -933,33 +933,38 @@ protected void unbindAndShutdown() {
         }
     }
 
-    protected void removeCurrentDialog() {
+    protected void removeCurrentDialog(Activity callerActivity) {
```
dominicmarmionswrve: I think we only need to remove the dialog if the owner activity of the dialog is finishing. You can find this out by calling dialog.getOwnerActivity().isFinishing(). If its nots finishing then don't remove the dialog. This way we don't need to add Activity to the lifecyle methods.

Sergio-Mira: isFinishing is not reliable when checked inside of onDestroy. It is only
reliable inside onPause as far as I know.

dominicmarmionswrve: Ok, I didn't know that. I've googled this but can't find anything to support it. You sure?

Sergio-Mira: The problem is that the activity that is calling onDestroy (the first one) is actually going away, so it will remove the current dialog.

dominicmarmionswrve: Yeah, but the first activity is not the owner of the dialog. The dialog should only be removed if the owner activity is finishing. dialog.getOwnerActivity().isFinishing().

Sergio-Mira: True, will test it and change it if it is reliable.

SwrveSDKDemo/src/com/swrve/sdk/demo/MainActivity.java
```diff
@@ -126,7 +126,15 @@ protected void onResume() {
         super.onResume();
         // Notify the SDK of activity resume.
         appAvailable = true;
-        swrve.onResume();
+        swrve.onResume(this);
+    }
+
+
+    @Override
+    protected void onLowMemory() {
```
dominicmarmionswrve: onLowMemory is public, not protected

SwrveSDKDemo/src/com/swrve/sdk/demo/MainActivity.java
```diff
@@ -142,6 +150,6 @@ protected void onDestroy() {
         super.onDestroy();
         // Notify the SDK of activity destroy.
         appAvailable = false;
-        swrve.onDestroy();
+        swrve.onDestroy(this);
     }
 }
```
dominicmarmionswrve: onDestroy, onPause, etc shouldn't have the 'this' param.

SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -937,29 +937,34 @@ protected void removeCurrentDialog() {
...
+                Activity dialogActivity = dialog.getOwnerActivity();
+                if (dialogActivity == null || dialogActivity.isFinishing()) {
+                    messageDisplayed = dialog.getMessage();
+                    // Remove reference to the SDK from the message
+                    messageDisplayed.setMessageController(null);
+                    lastMessageDestroyed = (new Date()).getTime();
+                    Activity activity = dialogActivity;
+                    if (activity == null) {
+                        activity = getActivityContext();
+                    }
```
dominicmarmionswrve: :+1:

dominicmarmionswrve:
:rocket:

PR #35, opened 2015-04-03
8073-null_context_after_ondestroy

dominicmarmionswrve:
only nullify activity_context when bind counter is zero
https://swrvedev.jira.com/browse/SWRVE-8073


dominicmarmionswrve:
@Sergio-Mira

Sergio-Mira:
:rocket:

PR #42, opened 2015-09-07
9733 Added configurable HTTP timeout

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-9733

### DIFF
--- a/SwrveCommonSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
+++ b/SwrveCommonSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
@@ -121,7 +121,7 @@ public abstract class SwrveConfigBase {
     private int defaultBackgroundColor = Color.TRANSPARENT;
 
     /**
-     * HTTP timeout used when contacting the Swrve APIs.
+     * HTTP timeout used when contacting the Swrve APIs, in milliseconds.
      */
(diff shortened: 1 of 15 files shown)


Sergio-Mira:

SwrveCommonSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
```diff
@@ -121,6 +121,11 @@
     private int defaultBackgroundColor = Color.TRANSPARENT;
 
     /**
+     * HTTP timeout used when contacting the Swrve APIs.
+     */
+    private int httpTimeout = 60000;
+
```
dominicmarmionswrve: Might be better to use constants in the future (and longs), especially for milliseconds. Can be error prone. eg:
private static final long SECOND = 1000;
private static final long httpTimeout = SECOND*60;

dominicmarmionswrve:
:rocket:

SwrveCommonSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
dominicmarmionswrve: Actually, better the constant is SECOND_MS to indicate its milliseconds.

PR #46, opened 2015-09-08
9759 Fix NPE when dialog is dismissed and the same message is displayed

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-9759


dominicmarmionswrve:
:rocket:

PR #49, opened 2015-09-08
8310 React to HTTP code instead of the existence of a body

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-8310


Sergio-Mira:
Retargeting against Android 4.0.2

PR #57, opened 2015-09-22
9842-init location campaigns

dominicmarmionswrve:
init location campaigns and make campaign matching code similar to ios implementation

https://swrvedev.jira.com/browse/SWRVE-9842

tests:
https://github.com/Swrve/swrve-sdk/pull/908
@Sergio-Mira

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/locationcampaigns/LocationCampaignFilter.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/locationcampaigns/LocationCampaignFilter.java
@@ -16,10 +16,9 @@ import com.swrve.sdk.locationcampaigns.model.LocationMessage;
 import com.swrve.sdk.locationcampaigns.model.LocationPayload;
 
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 public class LocationCampaignFilter extends NotificationFilterReceiver {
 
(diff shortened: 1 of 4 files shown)


SwrveSDK/src/main/java/com/swrve/sdk/locationcampaigns/LocationCampaignFilter.java
```diff
@@ -35,7 +34,8 @@
         try {
             wakeLock.acquire();
 
-            ((SwrveBase)(SwrveSDKBase.getInstance())).initLocationCampaigns(); // TODO: presumption for FA is that location campaigns have been refreshed from ABTS by starting the app. Fix for GA.
+            // TODO: presumption for FA is that location campaigns have been refreshed from ABTS by starting the app. Also that certain variables have been initialised. For FA hack this. For GA fix this.
```
Sergio-Mira: Is onCreate on the Application class called before this? if so we can move that initSDKForLocationCampaigns into the createInstance method

SwrveSDK/src/main/java/com/swrve/sdk/locationcampaigns/model/LocationCampaign.java
```diff
@@ -9,14 +9,14 @@
 
 public class LocationCampaign  {
 
-    private final int id;
+    private String id;
```
Sergio-Mira: Is this really a string coming from the server?

SwrveSDK/src/main/java/com/swrve/sdk/locationcampaigns/LocationCampaignFilter.java
dominicmarmionswrve: No, onCreate will not be called from application class.

We will be refactoring the init method of the sdk so that a lot of this code should be initialised in the createInstance. Not doing this for FA though.

SwrveSDK/src/main/java/com/swrve/sdk/locationcampaigns/model/LocationCampaign.java
dominicmarmionswrve: Its an int currently. It could be a UUID if we wanted to. The sdk can handle either. The LocationMessage id is a string also.

PR #61, opened 2015-10-05
New Youtube Iframe API player for better youtube video stability.

shanetmoore:
- New regexp matchers to detect youtube URL's and pull out the video ID.
- New YoutubeVideoView to replace the HtmlVideoView (in the future we may add a vimeo view and create these views at the fragment level)
- When a youtube video is detected, the Webview will load in a hardcoded HTML file in the assets directory which handles pulling in the
JS required for proper rendering of youtube video.

This is a HTML5 based player supported by youtube which is under active development and deprecations. It does not solve ALL of our existing
video issues but does feel like it gives us a more stable youtube experience.

This should fix::
- The width sizing issue (only 80% of screen width utilized and aligned left on some tablets)
- Some of the more annoying flickers
- Gives us a better error handling mechanism to hook into

https://swrvedev.jira.com/browse/SWRVE-9738 
https://swrvedev.jira.com/browse/SWRVE-9736 
https://swrvedev.jira.com/browse/SWRVE-9720

### DIFF
--- a/SwrveSDK/src/main/assets/youtubeapi.html
+++ b/SwrveSDK/src/main/assets/youtubeapi.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
-<body>
-<div id="player"></div>
+<body style = "padding: 0; margin: 0;">
+<div style = "width: 100%; height: 100%;" id="player"></div>
 
 <script>
       var tag = document.createElement('script');
--- a/SwrveSDK/src/main/assets/youtubeapi.html
+++ b/SwrveSDK/src/main/assets/youtubeapi.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<body>
+<div id="player"></div>
+
+<script>
+      var tag = document.createElement('script');
+
+      tag.src = "https://www.youtube.com/iframe_api";
+      var firstScriptTag = document.getElementsByTagName('script')[0];
+      firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
+
+      var player;
+      function onYouTubeIframeAPIReady() {
+        player = new YT.Player('player', {
+          height: 'VIDEO_HEIGHT_PLACEHOLDER',
+          width:  '100%',
+          videoId: 'VIDEO_ID_PLACEHOLDER',
+          playerVars: {
+            iv_load_policy: 3, /* Don't show annotations by default */
+            modestbranding: 1, /* Don't show the youtube branding */
+            showinfo: 0, /* Dont' show title and video info before the user starts playing (its visually noisy) */
+            rel: 0, /* Don't show the related videos list when the video ends */
+            frameborder: 0 /* No border around the video */
+          }
+        });
+      }
+    </script>
+</body>
+</html>
--- a/SwrveSDK/src/main/java/com/swrve/sdk/conversations/engine/model/Content.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/conversations/engine/model/Content.java
@@ -1,11 +1,20 @@
 package com.swrve.sdk.conversations.engine.model;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public class Content extends ConversationAtom {
+    public static final String YOUTUBE_VIDEO_ID_REGEX = "^https?://.*(?:youtu.be/|v/|u/\\w/|embed/|watch?v=)([^#&?]*).*$";
+
     protected String value;
     protected String height;
 
     public String getValue() {
         return value;
+    }
+
+    public void setValue(String value) {
+        this.value = value;
     }
 
     public String getHeight() {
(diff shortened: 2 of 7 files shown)


shanetmoore:
@Sergio-Mira Can you please take a look at this and tell me what you think?

Maybe try it out on your phone and see if you notice the difference. 

Details on the Youtube Iframe API are here https://developers.google.com/youtube/iframe_api_reference

The available paramters are here: https://developers.google.com/youtube/player_parameters?playerVersion=HTML5

PS we might be able to use the loading color as a selling point to product. Using the branded color as the loading bar color etc

SwrveSDK/src/main/assets/youtubeapi.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<body style = "padding: 0; margin: 0;">
+<div style = "width: 100%; height: 100%;" id="player"></div>
+
+<script>
```
Sergio-Mira: Could we move this to an API that we can upgrade without any changes to the SDK? Or a file on an S3 bucket

SwrveSDK/src/main/java/com/swrve/sdk/conversations/ui/video/WebVideoViewBase.java
```diff
@@ -1,43 +1,44 @@
...
+    protected Content model;
+    protected Context context;
+    protected ConversationFullScreenVideoFrame fullScreenContainer;
+    protected int width;
+
 
-    public HtmlVideoView(Context context, Content model, ConversationFullScreenVideoFrame fullScreenContainer) {
+    public WebVideoViewBase(Context context, Content model, ConversationFullScreenVideoFrame fullScreenContainer) {
         super(context);
+        this.context = context;
```
Sergio-Mira: Does the WebView not already contain a context variable? Can't see it used anywhere either

shanetmoore: I could use getContext() each time I need it?

Its used for getting assets in the subclass view

Sergio-Mira:
Code looks class, will give it a go on a device tomorrow!

SwrveSDK/src/main/assets/youtubeapi.html
shanetmoore: We could. BUT, There are a lot of side effects to think about in the case. 
Storage, Extra Loading times, Support for the file would need to be maintained across all SDK's

This file is really just a much nicer way of generating HTML and not needing to make big long Strings in the java file.

shanetmoore: Short version is we could but we'd need to be absolutely sure we want to.

Sergio-Mira: I like it is in a file, however, if youtube change their API then video would stop working on old SDKs I guess

shanetmoore:
Cool lemme know how it goes. I'm still testing out 1 or 2 things.

SwrveSDK/src/main/assets/youtubeapi.html
shanetmoore: Hmmm. The JS is loaded via web. So the latest Iframe API playe JS is whats used. Only the HTML is hardcoded.

There are certain things we want local and certain things we want remote. We'll just have to weigh the pros/cons of both and figure out which one we want to associate this with.

SwrveSDK/src/main/java/com/swrve/sdk/conversations/engine/model/Content.java
```diff
@@ -1,6 +1,11 @@
 package com.swrve.sdk.conversations.engine.model;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public class Content extends ConversationAtom {
+    public static final String YOUTUBE_VIDEO_ID_REGEX = "^https?://.*(?:youtu.be/|v/|u/\\w/|embed/|watch?v=)([^#&?]*).*$";
```
Sergio-Mira: I think on the dash we support only youtube.com, not youtu.be

Sergio-Mira:
:rocket: (wait for confirmation from product)

SwrveSDK/src/main/java/com/swrve/sdk/conversations/ui/video/WebVideoViewBase.java
shanetmoore: Do you want me to change this as listed above?

SwrveSDK/src/main/java/com/swrve/sdk/conversations/engine/model/Content.java
shanetmoore: Indeed. It didn't cost anything to add the youtu.be to the SDKS. So if this guy comes in https://swrvedev.jira.com/browse/SWRVE-9267 and we decide to support him, there'll be no extra Android work needed :)

So we don't have to officially support it but I wanted to make sure it was robust enough.

Sergio-Mira: Cool :dancer:

SwrveSDK/src/main/java/com/swrve/sdk/conversations/ui/video/WebVideoViewBase.java
Sergio-Mira: Yeah please

shanetmoore:
@Sergio-Mira Release date confirmed for December.

PR #65, opened 2015-10-12
Fix for watch? urls in video ID

shanetmoore:
unescaped ? character caused tests to fail and watch?= ids not to be picked up


shanetmoore:
@Sergio-Mira

Sergio-Mira:
:rocket:

PR #66, opened 2015-10-12
9530 android swrve logger

shanetmoore:
New toggleable SwrveLogger class which is a wrapper around the Android Logger. It allows users to disable logs through the SwrveConfig class (this has been requested once already)

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveLogger.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveLogger.java
@@ -9,7 +9,7 @@ public class SwrveLogger {
     private static final String LOG_TAG = "SWRVE";
 
     // Config/Active/Inactive
-    private static boolean isActive = true;
+    private static boolean isActive = BuildConfig.DEBUG;
 
     public static boolean isActive() {
         return SwrveLogger.isActive;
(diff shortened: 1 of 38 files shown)


shanetmoore:

SwrveSDK/src/main/java/com/swrve/sdk/SwrveLogger.java
```diff
@@ -0,0 +1,165 @@
...
+import android.util.Log;
+
+import com.swrve.sdk.SwrveLogger;
+
+public class SwrveLogger {
+
+    private static final String LOG_TAG = "SWRVE";
+
+    // Config/Active/Inactive
+    private static boolean isActive = true;
```
dominicmarmionswrve: Would BuildConfig.debug be a better default?
private static boolean isActive = BuildConfig.debug;

dominicmarmionswrve:
Could we also replace all the occurrences of printStackTrace with this new Logger?

shanetmoore:
Done and Done

shanetmoore:
The stack trace loggers might not actually be needed in some cases.

I'm not sure printing the stacktrace provides over logging them with Log.e. IE `Log.e(tag, message, exception)` vs `e.printStackTrace()`

dominicmarmionswrve:
:rocket:

PR #70, opened 2015-10-19
9477 QA user only reset first time

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-9477


Sergio-Mira:
@dominicmarmionswrve @shanetmoore Could you have a look? Will write a test soon

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -763,6 +763,8 @@ protected void loadCampaignsFromJSON(JSONObject json, JSONObject settings) {
 
             Map<Integer, String> campaignsDownloaded = null;
 
+            boolean wasQAUser = (qaUser != null);
```
shanetmoore: was?

shanetmoore:
:+1: Syntactically looks fine. Just that question about the wasQAUser?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
Sergio-Mira: I can change it but wanted to reflect that we are checking it was a QA user already before we might make it one after this refresh (on that same method)

shanetmoore: Ah makes sense now. Could be worthy of a comment.

Sergio-Mira: I will change it to be "wasQAUserLastRefresh" maybe?

dominicmarmionswrve: I prefer something like wasPreviouslyAQAUser.
Its not a nice name, so as Shane says, best to add a comment.

dominicmarmionswrve:
Think it look ok. But needs tests.

PR #71, opened 2015-10-22
Added SwrveLogger to branched code which was missing it.

dominicmarmionswrve:
Added SwrveLogger to branched code which was missing it.

@shanetmoore TIny one here please.


shanetmoore:
:rocket:

PR #72, opened 2015-10-28
10160 Inbox API (REVIEW ONLY)

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-10160

### DIFF
--- a/SwrveCommonSDK/build.gradle
+++ b/SwrveCommonSDK/build.gradle
@@ -4,7 +4,7 @@ apply plugin: 'maven'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion '21.1.2'
+    buildToolsVersion '23.0.1'
 
     defaultConfig {
         minSdkVersion 10
--- a/SwrveCommonSDK/build.gradle
+++ b/SwrveCommonSDK/build.gradle
@@ -24,7 +24,7 @@ android {
         }
     }
     lintOptions {
-        abortOnError true
+        abortOnError false
     }
 }
 
--- a/SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -1837,7 +1837,7 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
 
     @Override
     public List<SwrveBaseCampaign> getCampaigns() {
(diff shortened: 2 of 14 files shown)


Sergio-Mira:

SwrveCommonSDK/build.gradle
```diff
@@ -4,7 +4,7 @@ apply plugin: 'maven'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion '21.1.2'
+    buildToolsVersion '23.0.1'
```
shanetmoore: :+1: Be mindful of tests failing due to this though. We've had issues with some roboelectrics as a result of build and compile versions.

SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -1848,4 +1831,30 @@ public C getConfig() {
...
+                    result.add((SwrveConversationCampaign)campaign);
+                }
+            }
+        }
+        return result;
+    }
+
+    @Override
+    public boolean displayConversationCampaign(SwrveConversationCampaign campaign) {
+        if (campaign != null && campaign.getConversation() != null && conversationListener != null) {
```
shanetmoore: I have a question on the use case here. Do we stop campaigns being shown when its expired? Or should we filter them out of the list?

Sergio-Mira: That is a good question! I'll ask product, but it would be good yeah if we follow start and and date rules for these too.

SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -1848,4 +1831,30 @@ public C getConfig() {
...
+            // Display first message in the campaign
+            conversationListener.onMessage(campaign.getConversation());
+            return true;
+        } else {
+            Log.e(LOG_TAG, "No conversation listener available.");
```
dominicmarmionswrve: Campaign could be null or no conversation, so perhaps change error message?

dominicmarmionswrve:
:rocket:

SwrveCommonSDK/src/main/java/com/swrve/sdk/SwrveBase.java
Sergio-Mira: I've made our SDK check if the campaigns are still active before returning them in the new method

PR #76, opened 2015-11-04
10154 timezone information ab

shanetmoore:
We're not sure if the timezone info is to be sent through to the AB test server after all.

If it is then the work is done here. We'll also need to wait for the locations SDK code to be pulled out and moved around.


Sergio-Mira:
This is not needed in the end, but I like that you put stuff into constants. We can merge if you take the timezone info :+1:

shanetmoore:
@Sergio-Mira Done.

PR #81, opened 2015-11-06
10342 easier stack configuration

shanetmoore:
Can now call setStack which will make it easier for developers to switch between stacks and not have to specific content or event urls manually.

Also a utility method `config.useEUStack()` which will set everything up automatically.

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -117,8 +117,7 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
             saveUniqueUserId(resolvedContext, userId);
             SwrveLogger.i(LOG_TAG, "Your user id is: " + userId);
 
-            // Generate default urls for the given app id
-            config.setAppIdPrefix(appId);
+            config.setUseHttpsForEventsUrl(true);
 
             if (SwrveHelper.isNullOrEmpty(config.getLanguage())) {
                 this.language = SwrveHelper.toLanguageTag(Locale.getDefault());
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -118,7 +118,7 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
             SwrveLogger.i(LOG_TAG, "Your user id is: " + userId);
 
             // Generate default urls for the given app id
-            config.generateUrls(appId);
+            config.setAppIdPrefix(appId);
 
             if (SwrveHelper.isNullOrEmpty(config.getLanguage())) {
                 this.language = SwrveHelper.toLanguageTag(Locale.getDefault());
--- a/SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
@@ -10,7 +10,6 @@ import com.swrve.sdk.messaging.SwrveOrientation;
 
 import java.io.File;
 import java.net.MalformedURLException;
-import java.net.URI;
 import java.net.URL;
 import java.util.Locale;
 
@@ -89,9 +88,9 @@ public abstract class SwrveConfigBase {
 
 
     /**
-     * Current stack choice
+     * Current currentStack choice
      */
-    private SwrveStack stack = SwrveStack.US;
+    private SwrveStack currentStack = SwrveStack.US;
 
     /**
      * Session timeout time.
@@ -329,13 +328,13 @@ public abstract class SwrveConfigBase {
     }
 
     /**
-     * Based on the stack that has been chosen, return the correct prefix
+     * Based on the currentStack that has been chosen, return the correct prefix
      * @return
      */
     private String getStackPrefix(){
         String stackPrefix = "";
 
-        switch (stack){
+        switch (currentStack){
             case US:
                 // No prefixes
                 break;
@@ -389,7 +388,7 @@ public abstract class SwrveConfigBase {
         boolean generatingContentURL = eventOrContent.equalsIgnoreCase("content");
 
         // Set the appID prefix but only for supported stacks
-        boolean isAppIdSupportedStack  = (stack == SwrveStack.US || stack == SwrveStack.EU || stack == SwrveStack.ACTIVISION);
+        boolean isAppIdSupportedStack  = (currentStack == SwrveStack.US || currentStack == SwrveStack.EU || currentStack == SwrveStack.ACTIVISION);
         if (null != appIdPrefix && isAppIdSupportedStack){
             authority = Integer.toString(appIdPrefix) + ".";
         }
@@ -409,7 +408,7 @@ public abstract class SwrveConfigBase {
 
         String prefix = getStackPrefix();
 
-        // Set the stack prefix and host
+        // Set the currentStack prefix and host
         authority = authority + getStackPrefix() + endpoint;
         Uri uri = uriBuilder.scheme(scheme).authority(authority).build();
         URL newUrl = new URL(uri.toString());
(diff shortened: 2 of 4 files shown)


shanetmoore:
@Sergio-Mira

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
```diff
@@ -296,6 +330,154 @@ public SwrveConfigBase setDbName(String dbName) {
...
+        return stackPrefix;
+    }
+
+    /**
+     * Unless the user/developer has already set the content or event url, regenerate the urls based on config information.
+     *
+     * @return
+     * @throws MalformedURLException
+     */
+    private SwrveConfigBase regenerateURLS() {
```
dominicmarmionswrve: why the regenerateURLS? Why not just generateURLS? Same for regenerateUrlFor below.....

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
```diff
@@ -16,6 +18,20 @@
...
+     * Default Log Tag
+     */
+
+    private final String LOG_TAG = "SwrveConfigBase";
+
+    /**
+     * List of supported stacks
+     */
+    public enum SwrveStack{
+        EU, US, ACTIVISION, FEATURESTACK
```
dominicmarmionswrve: Don't think we should expose FEATURESTACK publically.

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
```diff
@@ -59,6 +76,23 @@
...
+    /**
+     * Optional Featurestack Number if using featurestack
+     */
+    private Integer featureStackNum = 0;
+
+
+    /**
+     * Current stack choice
+     */
+    private SwrveStack stack = SwrveStack.US;
```
dominicmarmionswrve: rename stack to currentStack?

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
shanetmoore: Originally there was another method called generateURLS which did something different. Will update these.

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
shanetmoore: Will do

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
shanetmoore: This is something I asked about and didn't get an answer.

What is the security concern in exposing featurestacks via api? Is there something damaging developers can do?

Also, if it isn't ok to use featurestacks, why is it ok to expose activision?

@robertclancy Can I grab your opinion on this?

Sergio-Mira: Activision is used by customers, featurestacks are not. We can add ease of featurestack config on the Test App UI.

Sergio-Mira: There were comments made on the JIRA about not adding FS.

SwrveSDKDemo/src/com/swrve/sdk/demo/DemoApplication.java
```diff
@@ -13,8 +18,23 @@
...
+
+        try {
+            config.useEuStack();
+            config.setStack(SwrveConfigBase.SwrveStack.US);
+            config.setAppIdPrefix(2132);
+            config.setUseHttpsForEventsUrl(true);
+            config.setEventsUrl(new URL("https://featurestack12-api.swrve.com")); // Calling these directly should lock them
+            config.setContentUrl(new URL("https://featurestack12-content.swrve.com"));
+            config.setUseHttpsForEventsUrl(false); // Should still be http
+            config.setStack(SwrveConfigBase.SwrveStack.US); // Should still be featurestack 12
```
dominicmarmionswrve: Mistaken commit?

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
Sergio-Mira: Can we move this into its own file? That way it is nicer when using it:
config(SwrvStack.EU)

SwrveSDKDemo/src/com/swrve/sdk/demo/DemoApplication.java
```diff
@@ -13,8 +18,23 @@
...
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
+        }
+
         try {
-            SwrveSDK.createInstance(this, YOUR_APP_ID, YOUR_API_KEY);
+            SwrveSDK.createInstance(this, 2132, "AYgy6V2uk6f3d46ShfJ", config);
```
Sergio-Mira: Remove pls

SwrveSDKDemo/src/com/swrve/sdk/demo/DemoApplication.java
shanetmoore: Yup. My bad

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
```diff
@@ -296,6 +329,158 @@ public SwrveConfigBase setDbName(String dbName) {
...
+    public Integer getAppIdPrefix() {
+        return appIdPrefix;
+    }
+
+    /**
+     *
+     * @param appIdPrefix
+     * @return
+     */
+    public SwrveConfigBase setAppIdPrefix(Integer appIdPrefix) {
```
dominicmarmionswrve: Don't think this should be public. Confusing to user.

dominicmarmionswrve:
My simplistic view of this work was that we'd just add a setter method that took an enum of which stack to use, and there would be no other code changes. If a user called setContentUrl or setEventsUrl, then this would implicitly override all settings. I think I prefer this approach.

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
robertclancy: should activision be in this list either? Are we allowed tell all of our customers that activision have their own stack?

shanetmoore: I saw the comments but in the interest of convenience in our SDK, I thought I'd add them in.
OK if featurestacks aren't allowed and thats final. I'll remove them.
I definitely think activision shouldn't be in that case. We have a single customer who uses it, not multiple customers.

As for your last comment, I'm not sure what you mean. Do you mean make the constructor accept a stack paramater?

shanetmoore: @robertclancy Just removed it now.

SwrveSDK/src/main/java/com/swrve/sdk/config/SwrveConfigBase.java
shanetmoore: So if a customers wants to add an app id prefix to their URL EG `https://2132.eu.api.com`, how should they do it?

dominicmarmionswrve: I don't think they should ever do that. They just need to init the sdk with appId and apiKey and we take care of the rest.  if they change the appId through this method they might not change the apiKey.......

dominicmarmionswrve: But if they do want to do this for a reason (?), then they should just call setContentUrl or setEventUrl which should override everything else.

shanetmoore:
I made these changes 
- As refactors to make the SwrveConfig cleaner
- To support all of our stacks
- To clean some already existing code around URL generation.

Post PR review, that simplistic approach is what this code now does.

shanetmoore:
@Sergio-Mira @dominicmarmionswrve Right guys, this is done and all the unhappy bits should have been removed.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -117,8 +117,7 @@ protected T init(final Activity activity) throws IllegalArgumentException {
             saveUniqueUserId(resolvedContext, userId);
             SwrveLogger.i(LOG_TAG, "Your user id is: " + userId);
 
-            // Generate default urls for the given app id
-            config.generateUrls(appId);
+            config.setUseHttpsForEventsUrl(true);
```
Sergio-Mira: Accidental?

shanetmoore: The generateURLS method is private.Calling setUseHttpsForEventsUrl will cause the urls to regenerate instead.

Should I make the generateURLS method public?

Sergio-Mira: But if I say I don't want to use https for events, and use that config, this will turn it into https.

shanetmoore: Doesn't this depend on the boolean value at the end.

IE setUseHttpsForEventsUrl(false) will set it to use http.

Does that answer your question?

Sergio-Mira: But this is inside the Swrve init right? Are you not overriding whatever value customers might provide?

PR #83, opened 2015-11-06
Fix for location of Android maven package

shanetmoore:
Looks like at some point the rake and gradle tasks stopped publishing aar and POM files. I've updated the gradle task to put them in the SwrveSDK/publish directory


shanetmoore:

Sergio-Mira:
:rocket:

PR #84, opened 2015-11-09
Package up the Javadocs and Sources jars into each individual flavor

shanetmoore:


shanetmoore:
@Sergio-Mira

Sergio-Mira:
:rocket:

PR #88, opened 2015-11-11
9530 swrve logger debug fix

shanetmoore:
Default to true instead of BuildDebug flag for Android SwrveLogger


shanetmoore:
@dominicmarmionswrve

dominicmarmionswrve:
:rocket:

PR #89, opened 2015-11-12
Updated EU configs to have correct eu prefix

shanetmoore:
SDK team originally thought the eu prefix was eu.api. It is actually eu-api.


shanetmoore:
@Sergio-Mira

Sergio-Mira:
:rocket:

PR #90, opened 2015-11-12
10336-android-validate-event-name

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-10336

@Sergio-Mira @shanetmoore please

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -169,7 +169,7 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
             String savedInstallTime = getSavedInstallTime();
             if (SwrveHelper.isNullOrEmpty(savedInstallTime)) {
                 // First time we see this user
-                event("Swrve.first_session");
+                _event("Swrve.first_session", null);
                 Date now = getNow();
                 userInstallTime = installTimeFormat.format(now);
             }
@@ -301,7 +301,7 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
     }
 
     protected void _event(String name) {
-        event(name, null);
+        _event(name, null);
     }
 
     protected void _event(String name, Map<String, String> payload) {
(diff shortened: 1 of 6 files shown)


SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -1372,7 +1372,9 @@ public void sessionEnd() {
     @Override
     public void event(String name) {
         try {
-            _event(name);
+            if(isValidEventName(name)) {
+                _event(name);
+            }
```
shanetmoore: Can we add a logger statement here to notify the user that they're event couldn't be sent due to it starting with Swrve?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -1381,12 +1383,29 @@ public void event(String name) {
...
             SwrveLogger.e(LOG_TAG, "Exception thrown in Swrve SDK", e);
         }
     }
 
+    private boolean isValidEventName(String name) {
+        List<String> restrictedNamesStartWith = new ArrayList<String>() {{
+            add("Swrve.");
+            add("swrve.");
+        }};
+
```
shanetmoore: Not a requirement but this is a handy one I use for list initialization: ` List<String> restrictedNamesStartWith = Arrays.asList("Swrve.", "swrve.");`

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
shanetmoore: Oh nvm I see the logger statement in the isValidEvent method

dominicmarmionswrve: :+1:

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
dominicmarmionswrve: Yup, knew that. Thought I'd go for this instead. :+1:

PR #95, opened 2015-11-16
10466-Reskin-demo_app

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-10466

@Sergio-Mira @shanetmoore Reskin of Demo app. Can had more features and functionality at later date.


SwrveSDKDemo/src/main/java/com/swrve/sdk/demo/ConfigContentFragment.java
```diff
@@ -0,0 +1,55 @@
...
+
+        String apikey = SwrveSDK.getApiKey();
+        TextView textViewApikey = (TextView) getView().findViewById(R.id.config_apikey);
+        textViewApikey.setText(apikey);
+
+        String userId = SwrveSDK.getUserId();
+        TextView textViewUserId = (TextView) getView().findViewById(R.id.config_userid);
+        textViewUserId.setText(userId);
+
+        //SwrveSDK.get
```
Sergio-Mira: Remove

SwrveSDKDemo/src/main/java/com/swrve/sdk/demo/MainActivity.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+        SwrveSDK.onResume(this);
+    }
+
+    @Override
+    public void onLowMemory() {
```
Sergio-Mira: Could you add 'onNewIntent' too?

Sergio-Mira:
:rocket:

SwrveSDKDemo/src/main/java/com/swrve/sdk/demo/MainActivity.java
dominicmarmionswrve: Do we have a SwrveSDK.onNewIntent? I don't see one......

Sergio-Mira: Maybe we should, we have processIntent(intent) instead

dominicmarmionswrve: OK, will have a think about that one and add as a new jira if we think we'll do it.

PR #99, opened 2015-11-27
Added on newIntent

Sergio-Mira:
Although we catch our target intents in onResume it makes sense too to have support for onNewIntent


Sergio-Mira:

SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
```diff
@@ -61,6 +61,8 @@
 
     void onLowMemory();
 
+    void onNewIntent(Intent intent);
```
dominicmarmionswrve, line 64: If there is only a google implementation of this then perhaps only add it to the google ISwrve subclass?

shanethmoore, line 64: Nice spot man! I missed that

PR #101, opened 2015-12-04
10130-refactor_android_init

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-10130
Refactored the initialisation of sdk. Removed the parsing of location campaigns from main sdk.

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -129,7 +129,7 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
 
             this.resourceManager = new SwrveResourceManager();
             if (preloadRandC) {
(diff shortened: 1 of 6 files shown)


SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -784,7 +753,6 @@ public void onResponse(RESTResponse response) {
 
                                     if (responseJson.has("location_campaigns")) {
```
Sergio-Mira: Are we using the cache at all now?

Sergio-Mira:
:rocket: !

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
dominicmarmionswrve: The location campaigns are saved locally to the sqlite db and we read directly from that. I'll add a note.

PR #104, opened 2015-12-18
Update build.gradle

aradomski:
Change dependency from + to specific version. Picked version 23.0.1 used SwrveSDKDemo to prevent unexpected situation. 
Picked Gson version 2.2.4 - newest version of 2.2+ line

### DIFF
--- a/SwrveSDK/build.gradle
+++ b/SwrveSDK/build.gradle
@@ -28,9 +28,9 @@ android {
 }
 
 dependencies {
-    compile 'com.google.code.gson:gson:2.2.+'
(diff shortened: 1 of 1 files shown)


shanetmoore:
Hi aradomski,

Thanks for taking the time to write this PR.

The + in build gradle dependencies versions is intentional. The + will use the latest by default.

We use + to be as flexible as possible with our customers integrations and if they are using these libraries in their own apps.

If you wish to use a specific version of the gson or app support libraries, you can specify them in your apps build.gradle

PR #106, opened 2016-01-07
10959-location-impression-events-bug

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-10959

@shanetmoore

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/QueuedEventsManager.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/QueuedEventsManager.java
@@ -87,7 +87,8 @@ public class QueuedEventsManager {
             }
 
             @Override
-            public void onException(Exception exp) {
+            public void onException(Exception ex) {
+                SwrveLogger.e(LOG_TAG, "Error posting batch of events.", ex);
             }
         });
     }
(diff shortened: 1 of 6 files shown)


shanethmoore:
Will developers have to declare this wakeful service in their manifests?

dominicmarmionswrve:
Similar to the ConversationActivity, it will be defined in our aar file, so they won't be required to define it in their manifest.

SwrveSDK/src/main/java/com/swrve/sdk/QueuedEventsManager.java
```diff
@@ -0,0 +1,94 @@
...
+                    deleteEvents = false;
+                    SwrveLogger.e(LOG_TAG, "Error sending events to Swrve: " + response.responseBody);
+                }
+
+                // Resend if we got a server error (5XX)
+                listener.onResponse(deleteEvents);
+            }
+
+            @Override
+            public void onException(Exception exp) {
```
shanethmoore: I know this was copy/paste but while we're in the neighbourhood, can we throw a SwrveLogger.e in here?

shanethmoore:
Awesome work man :+1: :rocket: :moon: Really happy we're finally pulling stuff out of SwrveIMP and SwrveBase

SwrveSDK/src/main/java/com/swrve/sdk/QueuedEventsManager.java
dominicmarmionswrve: Yes, will do, good catch.

PR #107, opened 2016-01-11
10959-Only send impression events from background service and make thread safe

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-10959

@shanetmoore Further changes to my original PR. These changes allow new events to be stored and only send those new events. 

Also made it thread safe.

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SendEventsManager.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SendEventsManager.java
@@ -3,37 +3,82 @@ package com.swrve.sdk;
 import com.swrve.sdk.config.SwrveConfigBase;
 import com.swrve.sdk.localstorage.ILocalStorage;
 import com.swrve.sdk.localstorage.MemoryCachedLocalStorage;
+import com.swrve.sdk.localstorage.SQLiteLocalStorage;
 import com.swrve.sdk.rest.IRESTClient;
 import com.swrve.sdk.rest.IRESTResponseListener;
 import com.swrve.sdk.rest.RESTResponse;
 
 import org.json.JSONException;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.Map;
 
-public class QueuedEventsManager {
+public class SendEventsManager {
 
-    protected static final String LOG_TAG = "SwrveSDK";
+    private static final String LOG_TAG = "SwrveSDK";
+    private static final Object lock = new Object();
 
     private final SwrveConfigBase config;
+    private final IRESTClient restClient;
     private final String userId;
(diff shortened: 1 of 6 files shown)


SwrveSDK/src/main/java/com/swrve/sdk/SendEventsManager.java
```diff
@@ -61,13 +106,13 @@ public void onResponse(boolean shouldDelete) {
...
             } catch (JSONException je) {
-                SwrveLogger.e(LOG_TAG, "Unable to generate event batch", je);
+                SwrveLogger.e(LOG_TAG, "Unable to generate event batch, and send events", je);
             }
         }
         return eventsSent;
     }
 
-    protected void postBatchRequest(String postData, final IPostBatchRequestListener listener) {
+    private void postBatchRequest(final String postData, final IPostBatchRequestListener listener) {
```
shanethmoore: Does it need to be final to be stored?

**
NVM I see the onException log

SwrveSDK/src/main/java/com/swrve/sdk/SwrveWakefulService.java
```diff
@@ -3,19 +3,20 @@
...
 import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Map;
 
 public class SwrveWakefulService extends IntentService {
 
     private static final String LOG_TAG = "SwrveWakeful";
     public static final String EXTRA_SEND_EVENTS = "swrve_wakeful_events";
 
+    private Swrve swrve = (Swrve) SwrveSDKBase.getInstance();
```
shanethmoore, line 18: How will this behave if they haven't initialized the sdk?

dominicmarmionswrve, line 18: If they haven't initialized the sdk, then SwrveWakefulService will  not have been called. This is only being used by the Locations sdk (so far).

shanethmoore:
This all looks good. Is this so only Location events are sending or does this stop us sending events when the user restarts the app (causing a session start)

Only other suggestion I'd have is changing the name of `public class SendEventsManager {` to SwrveEventsManager. This is just to keep it inline with the rest of our classes and also to keep logic or implementation details out of the name.  Totally up to yourself though.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveWakefulService.java
shanethmoore, line 18: ah, ok noted

dominicmarmionswrve:
Yes, this is so only the location events, or whatever is passed to storeAndSendEvents is actually sent. If these don't get sent in background (when no network), they'll get sent next time the app is started in the normal way.

Cool, will rename to SwrveEventsManager.

PR #109, opened 2016-01-13
10443 asset retrying

shanethmoore:
Assets will retry if they fail during synchronous download

New refactored mechanism for storing assets in separate class instead of SwrveImp

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
@@ -1,0 +1,69 @@
+package com.swrve.sdk;
+
+import java.io.File;
+import java.util.HashSet;
+import java.util.Set;
+
+public class SwrveAssetsManager {
+    private final String LOG_TAG = "SwrveAssetsManager";
+    private final String cdnRoot;
+    protected Set<String> assetsOnDisk;
+    private Set<String> failedAssets;
+    protected boolean assetsCurrentlyDownloading;
+    protected File imageCacheDir;
+
+
+    public SwrveAssetsManager(File imageCacheDir, String cdnRoot) {
+        this.imageCacheDir = imageCacheDir;
+        this.cdnRoot = cdnRoot;
+        this.assetsOnDisk = new HashSet<String>();
+        this.failedAssets = new HashSet<String>();
+        this.assetsCurrentlyDownloading = false;
+    }
+
+    public String getCDNUrlForAssetPath(String assetPath) {
+        return cdnRoot + assetPath;
+    }
+
+    public boolean isAssetsCurrentlyDownloading() {
+        return assetsCurrentlyDownloading;
+    }
+
+    public void setAssetsCurrentlyDownloading(boolean assetsCurrentlyDownloading) {
+        this.assetsCurrentlyDownloading = assetsCurrentlyDownloading;
+    }
+
+    public void addAsset(String asset) {
+        failedAssets.remove(asset);
+        if (containsAsset(asset)) {
+            SwrveLogger.w(LOG_TAG, "Asset already present in SwrveAssetManager");
+            return;
+        } else {
+            assetsOnDisk.add(asset);
+        }
+    }
+
+    public void addFailedAsset(String asset) {
+        failedAssets.add(asset);
+    }
+
+    public boolean containsAsset(String asset) {
+        return assetsOnDisk.contains(asset);
+    }
+
+    public void clearAssetsOnDisk(){
+        this.assetsOnDisk.clear();
+    }
+
+    public boolean hasFailedAssets() {
+        return failedAssets.size() > 0;
+    }
+
+    public Set<String> getFailedAssets(){
+        return this.failedAssets;
+    }
+
+    public Set<String> getAssetsOnDisk() {
+        return this.assetsOnDisk;
+    }
+}
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -26,7 +26,7 @@ import com.swrve.sdk.messaging.ISwrveMessageListener;
 import com.swrve.sdk.messaging.SwrveActionType;
 import com.swrve.sdk.messaging.SwrveBaseCampaign;
 import com.swrve.sdk.messaging.SwrveButton;
-import com.swrve.sdk.messaging.SwrveCampaign;
+import com.swrve.sdk.messaging.SwrveMessageCampaign;
 import com.swrve.sdk.messaging.SwrveConversationCampaign;
 import com.swrve.sdk.messaging.SwrveEventListener;
 import com.swrve.sdk.messaging.SwrveMessage;
@@ -115,6 +115,8 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
 
             // Find cache folder
             findCacheFolder(activity);
+
+			this.initAssetManager();
 
             // Open access to local storage
             openLocalStorageConnection();
@@ -805,7 +807,7 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
     @SuppressLint("UseSparseArrays")
     protected SwrveConversation _getConversationForEvent(String event) {
         SwrveConversation result = null;
-        SwrveCampaign campaign = null;
+        SwrveMessageCampaign campaign = null;
 
         Date now = getNow();
         Map<Integer, String> campaignReasons = null;
@@ -876,7 +878,7 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
     @SuppressLint("UseSparseArrays")
     protected SwrveMessage _getMessageForEvent(String event, SwrveOrientation orientation) {
         SwrveMessage result = null;
-        SwrveCampaign campaign = null;
+        SwrveMessageCampaign campaign = null;
 
         Date now = getNow();
         Map<Integer, String> campaignReasons = null;
@@ -898,8 +900,8 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
                 Iterator<SwrveBaseCampaign> itCampaign = campaigns.iterator();
                 while (itCampaign.hasNext()) {
                     SwrveBaseCampaign nextCampaign = itCampaign.next();
-                    if (nextCampaign instanceof SwrveCampaign) {
-                        SwrveMessage nextMessage = ((SwrveCampaign)nextCampaign).getMessageForEvent(event, now, campaignReasons);
+                    if (nextCampaign instanceof SwrveMessageCampaign) {
+                        SwrveMessage nextMessage = ((SwrveMessageCampaign)nextCampaign).getMessageForEvent(event, now, campaignReasons);
                         if (nextMessage != null) {
                             // Add to list of returned messages
                             availableMessages.add(nextMessage);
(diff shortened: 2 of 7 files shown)


shanethmoore:

shanethmoore:
No idea if we still want this or https://github.com/Swrve/swrve-sdk/pull/1061 to go into r4.2 but sure lemme know what you think

SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
```diff
@@ -0,0 +1,69 @@
+package com.swrve.sdk;
+
+import java.io.File;
+import java.util.HashSet;
+import java.util.Set;
+
+public class SwrveAssetsManager {
```
dominicmarmionswrve, line 7: +1 Nice!

SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
```diff
@@ -0,0 +1,69 @@
...
+    private final String LOG_TAG = "SwrveAssetsManager";
```
dominicmarmionswrve, line 8: Should we just use a standard LOG_TAG for everything? eg: "SwrveSDK"

SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public SwrveAssetsManager(File imageCacheDir, String cdnRoot) {
+        this.imageCacheDir = imageCacheDir;
+        this.cdnRoot = cdnRoot;
+        this.assetsOnDisk = new HashSet<String>();
+        this.failedAssets = new HashSet<String>();
+        this.assetsCurrentlyDownloading = false;
+    }
+
+    public String getCDNUrlForAssetPath(String assetPath) {
```
dominicmarmionswrve, line 24: rename to getCdnUrlForAssetPath is a little more readable?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveHelper.java
```diff
@@ -201,4 +204,8 @@ public static boolean hasFileAccess(String filePath) {
             return false;
         }
     }
+
+    protected static boolean checkPermissionGranted(Context context, String permission) {
```
dominicmarmionswrve, line 208: +1!

SwrveSDK/src/main/java/com/swrve/sdk/messaging/SwrveMessage.java
```diff
@@ -19,7 +19,7 @@
  * In-app message inside a campaign, with different formats.
  */
 public class SwrveMessage {
-    protected static final String LOG_TAG = "SwrveSDK";
+    protected static final String LOG_TAG = "SwrveMessage";
```
dominicmarmionswrve, line 22: Same question as I asked above....should we use one standard LOG_TAG everywhere....."SwrveSDK"?

SwrveSDK/src/main/java/com/swrve/sdk/messaging/SwrveMessage.java
```diff
@@ -30,13 +30,13 @@
     // Priority of the message
     protected int priority = 9999;
     // Parent in-app campaign
-    protected SwrveCampaign campaign;
+    protected SwrveMessageCampaign campaign;
```
dominicmarmionswrve, line 33: Nice rename of class! Could we rename it to SwrveCampaignMessage and also the corresponding one should be SwrveCampaignConversation? That way they appear beside it each other in file explorer and also in auto complete.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -645,7 +651,7 @@ protected void findCacheFolder(Activity activity) {
     }
 
     protected boolean checkPermissionGranted(Context context, String permission) {
-        return ActivityCompat.checkSelfPermission(context, permission) == PackageManager.PERMISSION_GRANTED;
+        return SwrveHelper.checkPermissionGranted(context, permission);
```
dominicmarmionswrve, line 654: Probably best to just call SwrveHelper.checkPermissionGranted directly instead of going through this method. Also, might aswell move requestPermissions helper to the SwrveHelper class too and keep them together.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public void setAssetsCurrentlyDownloading(boolean assetsCurrentlyDownloading) {
+        this.assetsCurrentlyDownloading = assetsCurrentlyDownloading;
+    }
+
+    public void addAsset(String asset) {
+        failedAssets.remove(asset);
+        if (containsAsset(asset)) {
+            SwrveLogger.w(LOG_TAG, "Asset already present in SwrveAssetManager");
```
dominicmarmionswrve, line 39: Should this log level be info instead of warning?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -970,36 +976,36 @@ private boolean supportsDeviceFilter(String requirement) {
...
-    protected SwrveCampaign loadCampaignFromJSON(JSONObject campaignData, Set<String> assetsQueue) throws JSONException {
-        return new SwrveCampaign((SwrveBase<?, ?>) this, campaignData, assetsQueue);
-    }
-
-    protected SwrveConversationCampaign loadConversationCampaignFromJSON(JSONObject campaignData, Set<String> assetsQueue) throws JSONException {
-        return new SwrveConversationCampaign((SwrveBase<?, ?>) this, campaignData, assetsQueue);
-    }
-
+    @RequiresPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE)
     protected boolean downloadAssetSynchronously(final String assetPath) {
```
dominicmarmionswrve, line 1007: We should move the downloadAssetSynchronously to the SwrveAssetManager also and rename it just downloadAsset.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -970,36 +976,36 @@ private boolean supportsDeviceFilter(String requirement) {
...
-                            assetsOnDisk.add(asset);
+                        if (success) {
+                            assetManager.addAsset(asset);
+                        } else {
+                            assetManager.addFailedAsset(asset);
                         }
                     }
+
+                    if (assetManager.hasFailedAssets()) {
+                        downloadAssets(assetManager.getFailedAssets());
```
dominicmarmionswrve, line 997: Am concerned this keeps retrying to download the asset in a loop thus draining the battery. It should only retry a finite amount of times before the app is started again. Unless I'm missing something?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -805,7 +807,7 @@ private boolean checkCampaignRules(int elementCount, String elementName, String
     @SuppressLint("UseSparseArrays")
     protected SwrveConversation _getConversationForEvent(String event) {
         SwrveConversation result = null;
-        SwrveCampaign campaign = null;
+        SwrveMessageCampaign campaign = null;
```
dominicmarmionswrve, line 810: Could we rename this variable to swrveMessageCampaign? (or swrveCampaignMessage if the class is renamed?)

dominicmarmionswrve, line 810: Is it strange that we use the SwrveMessageCampaign in this method for conversations? Am confused. (Sorry, I know this isn't directly part of your code changes)

Sergio-Mira, line 810: That variable isn't used for anything anymore if I am correct, so could be removed

SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
shanethmoore, line 24: Sure thing

SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
shanethmoore, line 8: The main reason I've been using different tags (usually with the SwrvePrefix or name of the class) is because it makes it easier to track down where a log is coming from. 

If they all have SwrveSDK as the tag then we have to search by the log message.

SwrveSDK/src/main/java/com/swrve/sdk/messaging/SwrveMessage.java
shanethmoore, line 33: Good idea

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
shanethmoore, line 654: :+1:

SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
shanethmoore, line 39: I wasn't sure. In theory this should never actually happen so It seems like a warning. But sure I'll retag it to I once dev kicks off again on this branch

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
shanethmoore, line 1007: I'd also like to move the filter existing files method and one or 2 others :)     Its really only downloadAssets threading mechanism that was tripping me up due to future method calls to autoShowMessages.

Theres still a bit of room for the asset manager to grow methinks

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
shanethmoore, line 997: Yeah this is where my logic skills were concerned. In theory this would run recursively until the stack ran out of space so a finite depth level would be essential IMO

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
shanethmoore, line 810: Its mostly because of the initial implementation. We brought in SwrveConversations as a copy of SwrveMessages and theres still some tech debt around it.

If the var isn't used however then I'll :hocho:

SwrveSDK/src/main/java/com/swrve/sdk/SwrveAssetsManager.java
dominicmarmionswrve, line 39: If it should never happen, then I guess warning is a good level to have it at.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
dominicmarmionswrve, line 997: Agreed.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
dominicmarmionswrve, line 1007: Yes, this would be a lovely place for all that, and would make it far easier to unit test. I would definitely encourage this! :)

dominicmarmionswrve:
Nice job on all these changes btw! Its definitely worth continuing with as its a massive improvement to the way the code is structured! (But next release, not 4.2) :beers:

Sergio-Mira:
:clap:

Sergio-Mira:
Please retarget to 4.3

PR #110, opened 2016-01-14
Add a gradle.properties file, set commonly used versions there, and reference those versions in any build.gradle in the module folders.

seaders:
Mixing and matching Android's design tools version doesn't play well *, so leaving an ambiguous version like `23.+` for `com.android.support:support-v4` on one module will give you your machine's latest version of that library, but `23.0.1` on another module will give you that exact version.  At runtime, if the `SwrveSDK` module was built with a different design tools version than `SwrveSDKDemo`, the app will instantly crash giving out about some missing, or incorrect files on the internal Android Dex classpath.
- http://stackoverflow.com/questions/34046525/didnt-find-class-android-support-v7-internal-widget-tintmanager-on-path-dexp
- http://stackoverflow.com/questions/34015427/fatal-exception-main-java-lang-noclassdeffounderror-android-support-v7-interna


dominicmarmionswrve:
But doesn't this lock us into forcing our customers to use a specific version of the support library then? My understanding is that if use 23.+ in our library, it won't exactly force the latest version, because the actual version used will be that of the App who uses the library and declares it like 23.0.1 in their gradle build file. If the App doesn't specifically use support library, then our library in this instance will use the latest.

We had a customer moan about this before, so would be good to find out for sure!

+1 on using gradle properties

seaders:
I've been hit with both sides of this, being locked into a specific version, or using `23.+`-like references, and realistically it's messy either way.  For this specific thing, I downloaded this sdk, opened `Android Studio` and updated my Android Tools bits that it told me to do, ran the `SwrveSDKDemo`, and because my new tools are now at v `23.1.1`, mismatch, and crash, so anyone else who did the same as me would have experienced the same.

I've just checked there, and using `gradle` references,

```
compile 'com.android.support:appcompat-v7:23.+'
compile 'com.android.support:design:23.+'
compile 'com.swrve.sdk.android:swrve-google:4.1.0'
```

Runs without a problem, so you're definitely right with that.

I've recommitted with using `23.+` across the board (specifically in the `SwrveSDKDemo` module).

Main thing is from using libraries like this, any time I've run in to "weird" problems, where there's not a lot of previous on places like StackOverflow, or solutions offered, taking the step to externalise gradle versions, and ensure each project is using the same version as each other, and the installed tools on your machine has been the solution numerous times.

Sergio-Mira:
Redone in another PR

PR #111, opened 2016-01-14
REVIEW ONLY: Separated Conversations into its own module

Sergio-Mira:
Separated Conversations into its own module to be used by the Unity SDK


dominicmarmionswrve:
@Sergio-Mira Awesome work on this! Looks great, and we can make it nicer in time. One thing maybe worth mentioning would be the possibility of making a Common library where we could keep things like SwrveLogger, etc, and perhaps the main ISwrve interface?

Sergio-Mira:
Redone in another pr

PR #113, opened 2016-01-20
process push notifications on init

wrbooth:
While doing push notification integration in my app, I noticed that push notification payloads were not getting processed when the notification launched the app.  It appears this logic was simply getting skipped  on launch (init) while correctly being executed onResume.  

This may not be the best placement for the afterBind() call, but it certainly fixes the bug in question in my application.  Please let me know what you think.


Sergio-Mira:
@wrbooth could it be that your Activity is configured to only have one instance? In that case onCreate() is not called with the intent.
We are adding onNewIntent to our SDK, just an alias for now for processIntent(intent). You could add it to your Activity's onNewIntent and call processIntent(Intent).

Let us know how it goes!

PR #114, opened 2016-01-25
11037 inbox port

shanethmoore:
Merge of inbox API from 4.0.\* to 4.2

@Sergio-Mira Can you have a look at this?

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
@@ -6,9 +6,11 @@ import android.content.Context;
 import com.swrve.sdk.config.SwrveConfigBase;
 import com.swrve.sdk.messaging.ISwrveCustomButtonListener;
 import com.swrve.sdk.messaging.ISwrveDialogListener;
+import com.swrve.sdk.messaging.SwrveBaseCampaign;
 import com.swrve.sdk.messaging.ISwrveInstallButtonListener;
 import com.swrve.sdk.messaging.ISwrveMessageListener;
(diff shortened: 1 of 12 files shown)


shanethmoore:
Original PR here for reference https://github.com/Swrve/swrve-android-sdk/pull/72

Sergio-Mira:
:rocket:

SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
```diff
@@ -110,4 +113,12 @@
     Context getContext();
 
     C getConfig();
+
+    List<SwrveBaseCampaign> getCampaigns();
+
+    List<SwrveBaseCampaign> getCampaigns(SwrveOrientation orientation);
```
dominicmarmionswrve, line 119: Do these methods only get inbox campaigns?

Sergio-Mira, line 119: Yes, only inbox

dominicmarmionswrve, line 119: Should we rename it to getInboxCampaigns?

Sergio-Mira, line 119: I was instructed to not use the word Inbox as this feature is not really inbox, no UI yet :S

dominicmarmionswrve, line 119: getCampaigns is the wrong name for this api, so I don't think we should use that.

PR #118, opened 2016-02-01
Send more than one push notification to QAUsers.

josealfonsomora:
From Campaigns/Push Notifications on Preview & Test we cand fire
onPushNotification listeners again and again to QA users.
With this code we don't send any tracking engagement if we are
a qa user.

This a good update for developers because we don't have to create
multiple campaigns to test the push notification integration.


Sergio-Mira:
Thanks @josealfonsomora we'll look into including it into the upcoming 4.3 release!

Sergio-Mira:
We've added code to allow QA pushes to be received multiple times in our latest SDK 4.7, let us know how it goes and open an Issue if you find any problems, thanks!

PR #119, opened 2016-02-01
Added extra logging to the restclient so we can map error to the class

shanethmoore:
Also added a log tag


shanethmoore:
@dominicmarmionswrve 

https://swrvedev.jira.com/browse/SWRVE-11187

dominicmarmionswrve:
:rocket:

PR #121, opened 2016-02-08
message_center_null_subject

dominicmarmionswrve:
If the subject in message center is null, then a blank string is assigned.


Sergio-Mira:
:rocket:

PR #125, opened 2016-02-24
Let Swrve use app's own GCM registrationId token

Nilhcem:
Idea is for an app that already uses GCM to send its own GCM registrationId token to Swrve instead of letting Swrve do the GCM registration and have its own token.

When initializing Swrve, an app should call the following:

``` java
SwrveConfig config = new SwrveConfig();
config.provideOwnGcmRegistrationId();
SwrveSDK.createInstance(context, APP_ID, API_KEY, config);
```

Then, the app must handle the GCM registration on its own (with `InstanceID.getInstance(context).getToken()`) and send the received token to Swrve with:

``` java
SwrveSDK.setGcmRegistrationId(registrationId);
```

The app should, of course, keep the `SwrveGcmIntentService` defined in its `AndroidManifest.xml`, and also have the `SwrveIdentifyPushesBroadcastListener` (as specified in your documentation). The `SwrveGcmInstanceIDListenerService` is no longer necessary, as it is the app's responsibility to deal with expired tokens and send back the new one to Swrve with `SwrveSDK.setGcmRegistrationId(registrationId);`.
### Why this PR?

We had some concurrency issues with GCM registrations: When our app is started (first launch), it registers to GCM, but Swrve also does this similar process. We noticed that, when two GCM registrations (even with different SENDER_ID) are done at the same time, the tokens we received are usually invalid (their status is "NotRegistered). When doing the 2 GCM registrations sequentially (e.g. when using some breakpoints to slow down the registration process making sure Swrve receives its token, then our app), the tokens are valid.
We did not modify the code, only slowed down the registration process.

Here, we only want only our app to perform the GCM registration, not Swrve. When the gcm registration succeeds, we send you our registration token.

We don't have GCM issues anymore when using this modified version of Swrve.

### DIFF
--- a/SwrveSDK/src/google/java/com/swrve/sdk/ISwrve.java
+++ b/SwrveSDK/src/google/java/com/swrve/sdk/ISwrve.java
@@ -19,4 +19,6 @@ public interface ISwrve extends ISwrveBase<ISwrve, SwrveConfig> {
     void processIntent(Intent intent);
 
     void onTokenRefreshed();
+
+    void setRegistrationId(String gcmRegistrationId);
 }
--- a/SwrveSDK/src/google/java/com/swrve/sdk/Swrve.java
+++ b/SwrveSDK/src/google/java/com/swrve/sdk/Swrve.java
@@ -36,6 +36,7 @@ public class Swrve extends SwrveBase<ISwrve, SwrveConfig> implements ISwrve {
         super(context, appId, apiKey, config);
     }
 
+    @Override
     public void onTokenRefreshed() {
         registerInBackground(getContext());
     }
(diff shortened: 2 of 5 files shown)


Sergio-Mira:
Thanks @Nilhcem for letting us know of those issues! We'll have a look and schedule it for an upcoming release

Nilhcem:
Thanks :+1:

Sergio-Mira:
We've added similar changes to these in our latest SDK 4.7:
`config.setPushRegistrationAutomatic(false);`
and then:
`SwrveSDK.setRegistrationId(registrationId);`
Let us know how it goes and reopen the issue if needed!

PR #134, opened 2016-03-07
11668_curved_button_units

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-11668

@adam-govan @Sergio-Mira

### DIFF
--- a/SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationButton.java
+++ b/SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationButton.java
@@ -3,6 +3,7 @@ package com.swrve.sdk.conversations.ui;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.res.ColorStateList;
+import android.content.res.TypedArray;
 import android.graphics.Color;
 import android.graphics.Paint;
 import android.graphics.RectF;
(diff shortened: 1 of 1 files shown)


SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationButton.java
```diff
@@ -39,10 +41,20 @@ public ConversationButton(Context context, ButtonControl model, int defStyle) {
         setEllipsize(TextUtils.TruncateAt.END);
     }
 
-    private void initBorderRadius() {
-        if(model.getBorderRadius() != null) {
-            borderRadius = Float.parseFloat(model.getBorderRadius());
-            // TODO convert from percent to correct units SWRVE-11668
+    private void initBorderRadius(Context context) {
+        if (model.getBorderRadius() != null) {
```
Sergio-Mira, line 45: Could we wrap this in a try catch and default to 0?

Sergio-Mira:
:rocket:

SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationButton.java
dominicmarmionswrve, line 45: It is already wrapped in a try/catch in ConversationFragment:
https://github.com/Swrve/swrve-android-sdk/blob/release-4_3/SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationFragment.java#L146-L157

If an exception occurs then the conversation is not shown, which we could argue is better than giving a default.

Sergio-Mira, line 45: Sounds good

PR #141, opened 2016-03-15
convert height of button to pixels and max radius is half that

dominicmarmionswrve:
convert height of button to pixels and max radius is half that

https://swrvedev.jira.com/browse/SWRVE-11668

@Sergio-Mira

### DIFF
--- a/SwrveSDKCommon/src/main/java/com/swrve/sdk/SwrveHelper.java
+++ b/SwrveSDKCommon/src/main/java/com/swrve/sdk/SwrveHelper.java
@@ -1,8 +1,10 @@
 package com.swrve.sdk;
 
+import android.content.Context;
 import android.os.Build;
 import android.util.Base64;
-import com.swrve.sdk.SwrveLogger;
+import android.util.DisplayMetrics;
+import android.util.TypedValue;
 
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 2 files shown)


SwrveSDKCommon/src/main/java/com/swrve/sdk/SwrveHelper.java
```diff
@@ -201,4 +203,10 @@ public static boolean hasFileAccess(String filePath) {
             return false;
         }
     }
+
+    public static float convertDipToPixels(Context context, float dipValue) {
+        DisplayMetrics metrics = context.getResources().getDisplayMetrics();
+        float pixelValue = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP, dipValue, metrics);
```
Sergio-Mira, line 209:
~~~ you can do return here directly

Sergio-Mira:
:rocket: !

SwrveSDKCommon/src/main/java/com/swrve/sdk/SwrveHelper.java
dominicmarmionswrve, line 209: I like to return a variable value so I can breakpoint at the return and see exactly the value that is being returned. Otherwise if you breakpoint at a return expression, you have to evaluate it.

PR #144, opened 2016-03-17
added getRegistrationId and made it public

artemtarassov:
getRegistrationId was present but to accessible from SwrveSDK. the access to this function is often required in order to work with the deviceToken for push notifications (e.g. by third party services).

### DIFF
--- a/SwrveSDK/src/google/java/com/swrve/sdk/ISwrve.java
+++ b/SwrveSDK/src/google/java/com/swrve/sdk/ISwrve.java
@@ -19,4 +19,6 @@ public interface ISwrve extends ISwrveBase<ISwrve, SwrveConfig> {
     void processIntent(Intent intent);
 
     void onTokenRefreshed();
+    
+    String getRegistrationId() ;
 }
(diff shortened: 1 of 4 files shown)


Sergio-Mira:
Thanks @artemtarassov Is the use-case that you want to register the token with the Swrve SDK and notify other third parties of that same token?
Unfortunately it is an async process, so it would also require a way to notify you that the value is ready.
We'll have a look into it

artemtarassov:
@Sergio-Mira basically this is the use-case. for example the Helpshift SDK requires the deviceToken to be present. regarding the asynch process, it would be great to come up with something like SwrveSDK.setDeviceTokenListener(...).

Sergio-Mira:
Our latest GCM SDK 4.7 allows it to start without querying for a registration id and there is a new method called setRegistrationId that lets you set the reg id to be used when sending pushes. We hope that solves the problems when integrating with other SDKs by setting their token to our SDK.

Let us know how it goes and reopen the issue if needed!

PR #146, opened 2016-03-17
[android] 11915 Remove printouts, and try/catch with error logging within close()

seaders:


seaders:
Rocket here too, @Sergio-Mira ?

Sergio-Mira:
:rocket:  (this has to be removed once we get stability back)

PR #153, opened 2016-03-24
Removed setLayerType, just was hiding the text.

dominicmarmionswrve:
Conversation text disappeared with this change. Removing it so its back to original state,
@Sergio-Mira please


Sergio-Mira:
:rocket:

adam-govan:
:rocket: :+1:

seaders:
late to the party, but I wanna rocket too.

:rocket:

PR #158, opened 2016-04-01
Moved the deeplinking to main Swrve sdk so metrics get processed

dominicmarmionswrve:
- Deeplinking occurs in main swrve sdk as opposed to gcmhandler
- Tidied up processIntent for gcm with additional logcats
- Moved methods from ActionBehaviours to common area SwrveIntentHelper

https://swrvedev.jira.com/browse/SWRVE-11961

@Sergio-Mira please


Sergio-Mira:
:rocket:

PR #162, opened 2016-04-14
12246 Added gson parsing of new triggers

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-12246

Added gson parsing of triggers

@Sergio-Mira @seaders @adam-govan


Sergio-Mira:
:rocket:

PR #165, opened 2016-04-19
12246 - updated with new json syntax for trigger conditions

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-12246

@Sergio-Mira please


Sergio-Mira:
🚀

PR #168, opened 2016-04-21
Shutdown one-task executors when tasks are finished

silin:
There is a very big problem with leaked threads in all versions of SWRVE including latest.
**Please, pay attention on this critical issue, because it affects all apps that use Swrve worldwide and easily reproducible on any devices.** Depending on client app and how it uses Swrve it can lead to OOM and ANR errors.
The problem in leaked threads. Here is example (I used Swrve demo app to reproduce it).
1. Open demo app 
2. Open DDMS and Threads tab
3. Observe
   demo_initial_start_threads
   Here you can see all threads that are used by the app. I made some investigation and added description for threads without names. All of them are from Swrve. Ok, this info is needed just as a starting point
4. Press HOME button and move app to background. Restore app from background. Repeat several times. I repeated this about 10-20 times to get following screen. In client apps you do not need these wierd steps, I just use them, because demo app has only one screen. For all other client apps things are event worst, because it happens on opening/closing every screen.
   demo_after_some_time
   As you can see we have much more threads in waiting state. I investigated Swrve internals and can make you sure that all of them will not be closed. So as a result on the number of such threads can grow indefinitely. On every screen opening Swrve adds 2-3 new threads to this list. So **every app that use Swrve affected**. Moreover, most interested in any app users are affected more than others, because they use the app more intensively and use same app process longer. As a result they have a huge probability to face OOM or ANR when the list of such threads consume almost all available memory. 
   In case of our app in some cases (when user use it several hours in a row) it is possible to have **thousands** of such waiting threads.

After investigation I found that Swrve use thread executors not very correctly in some cases. 
In a few words executors are not shutdowned, when the only task for what they are created is finished. And in this case such executors just wait for a new task indefinitely. But it is also caused by the fact that Swrve create new instance of such executors every time.

In my fix I just shutdown executors that are already finished their job (because next time for this task a new instance will be created). Also I verified that all tasks are still executed and Swrve works correctly after applying my changes.
Here is a result after same steps:
demo_after_fix

As you can see no more zomby threads in a list. They appear only for a short time and then dissappear from the list as they should do

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
@@ -738,11 +738,12 @@ abstract class SwrveImp<T, C extends SwrveConfigBase> {
      * Ensure that after SwrveConfig.autoShowMessagesMaxDelay milliseconds autoshow is disabled
      */
     protected void disableAutoShowAfterDelay() {
-        ScheduledExecutorService timedService = Executors.newSingleThreadScheduledExecutor();
+        final ScheduledExecutorService timedService = Executors.newSingleThreadScheduledExecutor();
         timedService.schedule(new Runnable() {
             @Override
             public void run() {
                 autoShowMessagesEnabled = false;
+                timedService.shutdownNow();
             }
         }, config.getAutoShowMessagesMaxDelay(), TimeUnit.MILLISECONDS);
     }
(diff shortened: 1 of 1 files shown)


Sergio-Mira:
Thanks @silin for the super explanation! We'll include your changes in the next upcoming release and add a note for reviews to make sure we close any new executor.

Sergio-Mira:
I've targeted the upcoming 4.4 here https://github.com/Swrve/swrve-android-sdk/pull/176 (and removed the shutdown from the scheduled executor, as the task inside itself is recurring and it is stopped with the app or when a new timer is created)

PR #176, opened 2016-04-27
12440 Shutdown short-lived executors

Sergio-Mira:
From:
https://github.com/Swrve/swrve-android-sdk/pull/168


Sergio-Mira:

dominicmarmionswrve:
Can we put all the shutdownNow's in a finally block? Even if it, the code in the try block is perfectly safe, it's a nice pattern to adapt.

dominicmarmionswrve:
🚀  if finally blocks added.

Sergio-Mira:
Moved to finally block, merging

dominicmarmionswrve:
Btw, you can do try/finally blocks without the "catch", which is what I meant.

Sergio-Mira:
Cool! Modified in 4.4

PR #178, opened 2016-04-28
add message lifecycle listener

rusty120:
We needed to add a lifecycle listener to enable pausing/resuming games when in-app message dialogs are shown/dismissed.


Sergio-Mira:
Hi @rusty120 thanks for the pull request. We'll have a look!

Sergio-Mira:
We are planning on moving to use an Activity for IAM in the upcoming version. That means there will be no need to have an extra listener and use the OS callbacks. Will notify if / when released.

dominicmarmionswrve:
@rusty120 The latest (4.5) sdk starts a new activity for in-app-messages. So the activity that started the in-app message activity will be paused and then resumed by the OS. This should 
provide a place to do additional pausing/resuming of a game.
https://developer.android.com/reference/android/app/Activity.html#onPause()
https://developer.android.com/reference/android/app/Activity.html#onResume()

Does this work for you?

rusty120:
Yup, that should work fine.

Thanks!

PR #179, opened 2016-04-28
12170_extra_plot_logging

dominicmarmionswrve:
Exposing a method called getCachedData(String userId, String key);
https://swrvedev.jira.com/browse/SWRVE-12170

@Sergio-Mira @seaders please


Sergio-Mira:
🚀

PR #181, opened 2016-04-29
7447 Check is bitmap already recycled

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-7447

Shouldn't be necessary in the copy I have of the android sources, but not sure of all versions.


dominicmarmionswrve:
🚀

PR #186, opened 2016-05-05
Read ad info from user storage entry, not shared, and write to the same

seaders:
Also get, and save isLimitAdTrackingEnabled parameter.

### DIFF
--- a/SwrveSDK/src/google/java/com/swrve/sdk/Swrve.java
+++ b/SwrveSDK/src/google/java/com/swrve/sdk/Swrve.java
@@ -63,8 +63,8 @@ public class Swrve extends SwrveBase<ISwrve, SwrveConfig> implements ISwrve {
         // Google Advertising Id logging enabled and Google Play services ready
         if (config.isGAIDLoggingEnabled() && isGooglePlayServicesAvailable()) {
             // Load previous value for Advertising ID
-            advertisingId = cachedLocalStorage.getSharedCacheEntry(SWRVE_GOOGLE_ADVERTISING_ID_CATEGORY);
-            String isAdvertisingLimitAdTrackingEnabledString = cachedLocalStorage.getSharedCacheEntry(SWRVE_GOOGLE_ADVERTISING_LIMIT_AD_TRACKING_CATEGORY);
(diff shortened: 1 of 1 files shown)


adam-govan:
👍  🚀

Sergio-Mira:
Can you add a description of the issue? Why was it moved to a device setting to a user one?

seaders:
I think it was always meant to be a user one, if you check the test of it,

https://github.com/Swrve/swrve-sdk/blob/add8fa5e96240a388a323a7ca37d1f7f556bddc9/android/SwrveSDKTests/src/test/java/com/swrve/sdk/LocationCampaignTest.java#L404

`swrve.cachedLocalStorage.setAndFlushSecureSharedEntryForUser(swrve.userId, SWRVE_GOOGLE_ADVERTISING_ID_CATEGORY, "some-advertisingId", swrve.getUniqueKey());`

It tested the user cached storage, not a device one.  If you check how cached things are exposed via SwrveCommon, 

https://github.com/Swrve/swrve-android-sdk/blob/release-4_3_1/SwrveSDKCommon/src/main/java/com/swrve/sdk/ISwrveCommon.java#L55

`String getCachedData(String userId, String key);`

and how that's implemented,

https://github.com/Swrve/swrve-android-sdk/blob/release-4_3_1/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java#L1705

`public String getCachedData(String userId, String key) {
        ILocalStorage localStorage = null;
        try {
            localStorage = createLocalStorage();
            return localStorage.getSecureCacheEntryForUser(userId, key, getUniqueKey());`

again, it uses user storage, not device.

If we want to store this as a device thing, then I'll revert / change back, but I went with what the tests were doing, and how it had been designed to be exposed, and the way it was used by the only place that was using it,

https://github.com/Swrve/swrve-sdk/blob/dc7bbd13209d6c4cd8e6379860295a938b6bf6be/android/SwrveLocationSDK/src/main/java/com/swrve/sdk/SwrvePlot.java#L130

`String advertisingId = SwrveCommon.getInstance().getCachedData(userId, ISwrveCommon.SWRVE_GOOGLE_ADVERTISING_ID_CATEGORY);`

Which again seems to consider this should have been stored as per user storage.

PR #188, opened 2016-05-13
10310 Conversation priorities

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-10310


SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -876,13 +879,23 @@ protected SwrveConversation _getConversationForEvent(String event) {
...
+                                    // remove those from being candidates
+                                    candidateConversations.clear();
+                                }
+                                minPriority = nextConversation.getPriority();
+                                candidateConversations.add(nextConversation);
+                            }
                         }
                     }
                 }
-                if (availableConversations.size() > 0) {
```
dominicmarmionswrve, line 856: Is the variable availableConversations needed anymore?

Sergio-Mira, line 856: It is to iterate over them and specify in the QA logs why they were not shown

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -868,6 +868,9 @@ protected SwrveConversation _getConversationForEvent(String event) {
             }
             synchronized (campaigns) {
                 List<SwrveConversation> availableConversations = new ArrayList<SwrveConversation>();
```
dominicmarmionswrve, line 844: availableConversations variable is redundant now I think

Sergio-Mira, line 844: Its not, one list has the highest priority messages / conversations where we will pick randomly, the other has all of the campaign messages / conversations that were ready to be shown, but were not because the other was chosen.

SwrveSDK/src/main/java/com/swrve/sdk/conversations/SwrveConversation.java
```diff
@@ -24,6 +24,8 @@
     protected transient SwrveBase<?, ?> swrve;
     // Parent in-app campaign
     protected transient SwrveConversationCampaign campaign;
+    // Priority of the conversation
+    protected int priority = 9999;
```
dominicmarmionswrve: I'd put this in the SwrveBaseConversation, alongside the id/name/pages variables. Also move the getter there, but do not provide a setter for this.

Why the default value of 9999?

SwrveSDK/src/main/java/com/swrve/sdk/conversations/SwrveConversation.java
```diff
@@ -113,4 +119,15 @@ public ConversationPage getPageForControl(ControlBase control) {
...
     }
+
+    /**
+     * @return the message priority.
+     */
+    public int getPriority() {
+        return priority;
+    }
+
+    public void setPriority(int priority) {
```
dominicmarmionswrve: We shouldn't provide a setter for this.

dominicmarmionswrve: Also, the setters in the base class (setId/setName/setPages) should be removed for same reason.

dominicmarmionswrve:
I'd remove the setPriority method. Otherwise 🚀

PR #189, opened 2016-05-16
null checks on event name and payload

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-12247

@Sergio-Mira please


Sergio-Mira:
🚀

PR #191, opened 2016-05-16
added colour for icon in material notifications

dominicmarmionswrve:
@Sergio-Mira small addition here please......:)


Sergio-Mira:
🚀

PR #195, opened 2016-05-25
10650 IAM use Activity instead of Dialog

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-10650

Will make it not compatible with 4.4.0 if the app uses the dialog listener (removed) or they have their own rendering code (tiny changes).


SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -179,15 +184,35 @@ protected T init(final Activity activity) throws IllegalArgumentException {
...
-                            if (SwrveBase.this.activityContext != null) {
-                                final Activity activity = SwrveBase.this.activityContext.get();
-                                if (activity == null) {
-                                    SwrveLogger.e(LOG_TAG, "Can't display a message with a non-Activity context");
+                        public void onMessage(final SwrveMessage message) {
+                            // Start a Conversation activity to display the campaign
+                            if (SwrveBase.this.context != null) {
+                                final Context ctx = SwrveBase.this.context.get();
+                                if (ctx == null) {
+                                    SwrveLogger.e(LOG_TAG, "Can't display a conversation without a context");
```
dominicmarmionswrve: Wrong message - remove the word  "conversation"

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -179,15 +184,35 @@ protected T init(final Activity activity) throws IllegalArgumentException {
                 // Add custom message listener
                 if (messageListener == null) {
                     setMessageListener(new ISwrveMessageListener() {
-                        public void onMessage(final SwrveMessage message, boolean firstTime) {
-                            if (SwrveBase.this.activityContext != null) {
-                                final Activity activity = SwrveBase.this.activityContext.get();
-                                if (activity == null) {
-                                    SwrveLogger.e(LOG_TAG, "Can't display a message with a non-Activity context");
+                        public void onMessage(final SwrveMessage message) {
```
dominicmarmionswrve: Can we reduce the number of lines in the init method and refactor so that this logic is in its own method. Move the whole setMessageListener.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -179,15 +184,35 @@ protected T init(final Activity activity) throws IllegalArgumentException {
...
                                     return;
                                 }
-                                // Run code on the UI thread
-                                activity.runOnUiThread(new DisplayMessageRunnable(SwrveBase.this, activity, message, firstTime));
+
+                                Intent intent = new Intent(ctx, SwrveInAppMessageActivity.class);
+                                intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                                intent.putExtra("inAppMessage", message);
+                                intent.putExtra("hideToolbar", config.isHideToolbar());
+                                intent.putExtra("minSampleSize", config.getMinSampleSize());
```
dominicmarmionswrve: Don't think we should serialize any of these things. The config is accessible via the SwrveSDK. The SwrveMessage could be accessed in a protected way also via the singleton.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -251,6 +273,64 @@ public void run() {
         return (T) this;
     }
 
+    private void processIAMMessage(SwrveMessage message, String eventType, Bundle bundle) {
```
dominicmarmionswrve: Does this need to happen here? Do it in the SwrveInAppMessageActivity class?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
Sergio-Mira: I didn't want to make IAM dependent on people using the singleton. Some apps do not use it and it would break when upgrading.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
Sergio-Mira: That is the whole point, to not use the singleton instance but rather communicate with the instance that launched the IAM. This is the Android way rather than using static instances.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
dominicmarmionswrve: Fair point actually. There's no need for the dependency if its not necessary, or can be achieved with no hassle! 👍

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
dominicmarmionswrve: Cool! Its worth pulling this logic out into its own Handler class.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -251,6 +273,64 @@ public void run() {
...
+        if (eventType.equals("impression")) {
```
dominicmarmionswrve: use enum?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -1052,50 +1016,6 @@ protected void unbindAndShutdown(Activity callerActivity) {
         }
     }
 
-    protected void removeCurrentDialog(Activity callerActivity) {
```
dominicmarmionswrve, line 1055: So happy this is gone!

SwrveSDK/src/main/java/com/swrve/sdk/messaging/SwrveInAppCampaign.java
```diff
@@ -23,7 +23,7 @@
 /**
  * Swrve campaign containing an in-app message targeted to the current device and user id.
  */
-public class SwrveCampaign extends SwrveBaseCampaign {
+public class SwrveInAppCampaign extends SwrveBaseCampaign {
```
dominicmarmionswrve, line 27: 👍  on the name change, perhaps SwrveInAppMessageCampaign instead?

dominicmarmionswrve:
We shouldn't be making all those objects serializable. Why are they?

SwrveSDK/src/main/java/com/swrve/sdk/messaging/SwrveMessage.java
```diff
@@ -17,24 +18,24 @@
 /**
  * In-app message inside a campaign, with different formats.
  */
-public class SwrveMessage {
+public class SwrveMessage implements Serializable {
     protected static final String LOG_TAG = "SwrveSDK";
 
-    protected ISwrveCampaignManager campaignManager;
+    protected transient ISwrveCampaignManager campaignManager;
```
dominicmarmionswrve: yuck

Sergio-Mira:
The message is passed to the activity, so it needs to be serializable

SwrveSDK/src/main/java/com/swrve/sdk/messaging/ui/SwrveInAppMessageActivity.java
```diff
@@ -0,0 +1,77 @@
...
+        super.onCreate(savedInstanceState);
+        Intent intent = getIntent();
+        if (intent != null) {
+            Bundle extras = intent.getExtras();
+            if (extras != null) {
+                this.message = (SwrveMessage) extras.getSerializable("inAppMessage");
+                this.hideToolbar = extras.getBoolean("hideToolbar", true);
+                this.minSampleSize = extras.getInt("minSampleSize", 1);
+                this.messenger = (Messenger)extras.get("messenger");
+            }
```
dominicmarmionswrve, line 58: else exit?

SwrveSDK/src/main/java/com/swrve/sdk/messaging/view/SwrveMessageView.java
```diff
@@ -1,250 +1,341 @@
...
     }
 
-    /**
-     * Set the message show animation.
-     *
-     * @param showAnimation
-     */
-    public void setShowAnimation(int showAnimation) {
-        this.showAnimation = showAnimation;
+    private void notifyOfInstallButtonPress(SwrveButton button, String appInstallLink) {
```
dominicmarmionswrve: These notify methods belong in their own class as they have their own set of logic and rules which can be far easier unit tested as a separate thing.

SwrveSDK/src/main/java/com/swrve/sdk/messaging/ui/SwrveInAppMessageActivity.java
Sergio-Mira, line 58: There is a finish() when message is null at the bottom

SwrveSDK/src/main/java/com/swrve/sdk/messaging/view/SwrveMessageView.java
```diff
@@ -1,250 +1,341 @@
...
 
-    public SwrveMessageView(Context context, SwrveMessage message, SwrveMessageFormat format, boolean firstTime, int rotation, int minSampleSize) throws SwrveMessageViewBuildException {
-        this(context, message, format, null, null, firstTime, rotation, minSampleSize);
-    }
+    // Bitmap cache
+    protected Set<WeakReference<Bitmap>> bitmapCache;
 
-    public SwrveMessageView(Context context, SwrveMessage message, SwrveMessageFormat format, ISwrveInstallButtonListener installButtonListener,
-                            ISwrveCustomButtonListener customButtonListener, boolean firstTime, int rotation, int minSampleSize) throws SwrveMessageViewBuildException {
+    public SwrveMessageView(Context context, Messenger messenger, SwrveMessage message, SwrveMessageFormat format, int minSampleSize) throws SwrveMessageViewBuildException {
```
dominicmarmionswrve: Could we move all the messenger and click listener logic to the activity class? Am not sure it belongs here.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
dominicmarmionswrve: Ok, am going back on this. Making everything serializable is not good. (Should be parcelable anyway, but we shouldn't do that either). IAM should be dependent on the singleton. The whole sdk, to my knowledge, is anyways.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -176,21 +179,9 @@ protected T init(final Activity activity) throws IllegalArgumentException {
...
-                                if (activity == null) {
-                                    SwrveLogger.e(LOG_TAG, "Can't display a message with a non-Activity context");
-                                    return;
-                                }
-                                // Run code on the UI thread
-                                activity.runOnUiThread(new DisplayMessageRunnable(SwrveBase.this, activity, message, firstTime));
-                            }
-                        }
-                    });
+                    setDefaultMessageListener();
```
dominicmarmionswrve, line 183: 👍

SwrveSDK/src/main/java/com/swrve/sdk/messaging/ui/SwrveInAppMessageActivity.java
```diff
@@ -0,0 +1,113 @@
...
+            try {
+                // Create view and add as root of the activity
+                SwrveMessageView view = new SwrveMessageView(this, message,
+                        format, minSampleSize, defaultBackgroundColor);
+                setContentView(view);
+                if(savedInstanceState == null) {
+                    notifyOfImpression(message, format);
+                }
+            } catch (SwrveMessageViewBuildException e) {
+                e.printStackTrace();
```
dominicmarmionswrve: Use SwrveLogger

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -251,6 +239,43 @@ public void run() {
...
+                        SwrveLogger.e(LOG_TAG, "Can't display a in-app message without a context");
+                        return;
+                    }
+
+                    Intent intent = new Intent(ctx, SwrveInAppMessageActivity.class);
+                    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                    intent.putExtra("inAppMessage", message);
+                    intent.putExtra("hideToolbar", config.isHideToolbar());
+                    intent.putExtra("minSampleSize", config.getMinSampleSize());
+                    intent.putExtra("defaultBackgroundColor", config.getDefaultBackgroundColor());
```
dominicmarmionswrve: Use constants for all those values: inAppMessage, hideToolbar, minSampleSize, defaultBackgroundColor

SwrveSDK/src/main/java/com/swrve/sdk/messaging/view/SwrveMessageView.java
```diff
@@ -1,251 +1,312 @@
...
+                    // Position
+                    RelativeLayout.LayoutParams lparams = new RelativeLayout.LayoutParams(backgroundImage.getWidth(), backgroundImage.getHeight());
+                    lparams.leftMargin = button.getPosition().x;
+                    lparams.topMargin = button.getPosition().y;
+                    lparams.width = backgroundImage.getWidth();
+                    lparams.height = backgroundImage.getHeight();
+                    buttonView.setLayoutParams(lparams);
+                    buttonView.setImageBitmap(imageBitmap);
+                    buttonView.setScaleType(ScaleType.FIT_XY);
+                    buttonView.setOnClickListener(new OnClickListener() {
```
dominicmarmionswrve, line 197: This logic does not belong in the View component. Can it be moved to the Activity?

SwrveSDK/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="Theme.InAppMessage" parent="android:Theme">
```
dominicmarmionswrve: Should this be a theme if its configurable? Would be nicer if all the logic for it was in one spot.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -251,6 +239,43 @@ public void run() {
...
+                if (SwrveBase.this.context != null) {
+                    final Context ctx = SwrveBase.this.context.get();
+                    if (ctx == null) {
+                        SwrveLogger.e(LOG_TAG, "Can't display a in-app message without a context");
+                        return;
+                    }
+
+                    Intent intent = new Intent(ctx, SwrveInAppMessageActivity.class);
+                    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                    intent.putExtra("inAppMessage", message);
```
dominicmarmionswrve: I really don't think we should be serializing a SwrveMessage to the activity. And I don't like that everything is serializable as a result. And also transient! It smells of poor design.

All those SwrveMessages are cached in SwrveImp someplace? Is there an id the activity can receive and look up directly in SwrveImp?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -251,6 +239,43 @@ public void run() {
...
+                    intent.putExtra("hideToolbar", config.isHideToolbar());
+                    intent.putExtra("minSampleSize", config.getMinSampleSize());
+                    intent.putExtra("defaultBackgroundColor", config.getDefaultBackgroundColor());
+                    Handler iamHandler = new Handler(new Handler.Callback() {
+                        @Override
+                        public boolean handleMessage(Message msg) {
+                            try {
+                                Bundle bundle = msg.peekData();
+                                String eventType = bundle.getString("eventType");
+                                processIAMMessage(message, eventType, bundle);
```
dominicmarmionswrve: Why does everything have to get processed back here in SwrveImp? Its just events being sent, right? Can the Activity class not process these events itself?

Sergio-Mira: It does a bit more than just send events. It is here to not be dependent on the shared instance, which is there for ease of use for customers and for last resort for us, but if we can avoid it we should.

dominicmarmionswrve: Nope. Completely disagree with that. The shared instance is not there just for customers. Otherwise the method SwrveSDK.createInstance would return void.

SwrveSDK/src/main/res/values/themes.xml
Sergio-Mira: Only the ability to hide the title is configurable

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -1030,19 +1008,18 @@ public void run() {
         }
     }
 
-    protected void noMessagesWereShown(String event, String reason) {
-        SwrveLogger.i(LOG_TAG, "Not showing message for " + event + ": " + reason);
-        if (qaUser != null) {
-            qaUser.triggerFailure(event, reason);
+    public Context getContext() {
```
dominicmarmionswrve: make protected?

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -1368,7 +1301,7 @@ public void run() {
         }
     }
 
-    protected class DisplayMessageRunnable implements Runnable {
+    /*protected class DisplayMessageRunnable implements Runnable {
```
dominicmarmionswrve: remove?

SwrveSDK/src/main/java/com/swrve/sdk/messaging/SwrveMessage.java
```diff
@@ -128,21 +123,16 @@ protected void setCacheDir(File cacheDir) {
...
     }
 
     private SwrveMessageFormat createMessageFormat(SwrveMessage swrveMessage, JSONObject messageFormatData) throws JSONException {
-        int defaultBackgroundColor = campaignManager.getConfig().getDefaultBackgroundColor();
-        return new SwrveMessageFormat(swrveMessage, messageFormatData, defaultBackgroundColor);
-    }
-
-    public ISwrveCampaignManager getCampaignManager() {
-        return campaignManager;
+        return new SwrveMessageFormat(swrveMessage, messageFormatData);
```
dominicmarmionswrve: No need for this method now. Just call new SwrveMessageFormat wherever its needed.

SwrveSDK/src/main/java/com/swrve/sdk/messaging/ui/SwrveInAppMessageActivity.java
```diff
@@ -0,0 +1,153 @@
...
+    private boolean hideToolbar = false;
+    private int minSampleSize;
+    private int defaultBackgroundColor;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        sdk = (SwrveBase)SwrveSDK.getInstance();
+        if (sdk != null) {
```
dominicmarmionswrve: Neater to do this:
if(sdk==null) {
// print error message
finish();
return;
}

SwrveSDK/src/main/java/com/swrve/sdk/messaging/ui/SwrveInAppMessageActivity.java
```diff
@@ -0,0 +1,153 @@
...
+            try {
+                // Create view and add as root of the activity
+                SwrveMessageView view = new SwrveMessageView(this, message,
+                        format, minSampleSize, defaultBackgroundColor);
+                setContentView(view);
+                if(savedInstanceState == null) {
+                    notifyOfImpression(format);
+                }
+            } catch (SwrveMessageViewBuildException e) {
+                e.printStackTrace();
```
dominicmarmionswrve: Use SwrveLogger

SwrveSDK/src/main/java/com/swrve/sdk/messaging/view/SwrveMessageView.java
```diff
@@ -1,251 +1,312 @@
...
     }
 
-    private void dismissView() {
-        final ViewParent parent = getParent();
-        if (containerDialog == null) {
-            removeView(parent);
-        } else {
-            containerDialog.dismiss();
-        }
+    protected SwrveImageView createSwrveImage(Context context) {
```
dominicmarmionswrve: Not sure this method provides any functionality?

SwrveSDK/src/main/java/com/swrve/sdk/messaging/view/SwrveMessageView.java
```diff
@@ -1,251 +1,312 @@
...
+        return new SwrveImageView(context);
     }
 
-    protected void removeView(ViewParent parent) {
-        if (parent != null) {
-            ((ViewGroup) parent).removeView(SwrveMessageView.this);
-        }
-        destroy();
+    protected SwrveButtonView createSwrveButton(Context context, SwrveActionType type) {
```
dominicmarmionswrve: Not sure this method provides any functionality?

SwrveSDK/src/main/java/com/swrve/sdk/messaging/view/SwrveMessageView.java
```diff
@@ -1,251 +1,312 @@
...
+                    }
+                    weakBitmap = null;
+                }
+
+                bitmapCache.clear();
+                bitmapCache = null;
             }
-        } catch (Exception e) {
-            SwrveLogger.e(LOG_TAG, "Error while processing first impression", e);
+            System.gc();
```
dominicmarmionswrve, line 294: System.gc() .....why?

Sergio-Mira, line 294: Force liberate bitmaps. Depending on the API they are not tied to the Java object and might not be GCed in time for another IAM to show and boom, OutOfMemoryException

dominicmarmionswrve, line 294: OK, calling System.gc() is generally bad practice and shouldn't be relied on. I guess it can do no harm though. 👍

SwrveSDK/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dominicmarmionswrve, line 1: rename the file to swrve__themes.xml?

SwrveSDK/src/main/AndroidManifest.xml
```diff
@@ -16,6 +16,12 @@
             android:theme="@style/Theme.Conversations">
         </activity>
 
+        <activity
+            android:name=".messaging.ui.SwrveInAppMessageActivity"
+            android:launchMode="singleTask"
```
dominicmarmionswrve: 👍

SwrveSDK/src/main/java/com/swrve/sdk/messaging/ISwrveMessageListener.java
```diff
@@ -10,8 +10,6 @@
      * This method is invoked when a message should be shown in your app.
      *
      * @param message   message to be shown.
-     * @param firstTime indicates if this message was already showing and the app
-     *                  rotated.
      */
-    void onMessage(SwrveMessage message, boolean firstTime);
```
dominicmarmionswrve, line 16: Deprecate this method?

SwrveSDK/src/main/java/com/swrve/sdk/messaging/SwrveImage.java
```diff
@@ -3,10 +3,12 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.io.Serializable;
+
 /**
  * In-app message background image.
  */
-public class SwrveImage extends SwrveWidget {
+public class SwrveImage extends SwrveWidget implements Serializable {
```
dominicmarmionswrve: I can't see where/why SwrveImage, SwrveCampaignState, SwrveButton need to be Serializable? Can you show me

Sergio-Mira: Not anymore, will remove. It was when using Intent

SwrveSDK/src/main/java/com/swrve/sdk/messaging/SwrveMessage.java
```diff
@@ -20,39 +19,35 @@
...
-    protected ISwrveCampaignManager campaignManager;
     // Identifies the message in a campaign
     protected int id;
     // Name of the message
     protected String name;
     // Priority of the message
     protected int priority = 9999;
     // Parent in-app campaign
-    protected SwrveCampaign campaign;
+    protected transient SwrveInAppCampaign campaign;
```
dominicmarmionswrve: Is transient needed anymore?

SwrveSDK/src/main/java/com/swrve/sdk/messaging/ui/SwrveInAppMessageActivity.java
```diff
@@ -0,0 +1,155 @@
...
+                } else {
+                    setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT);
+                }
+            }
+        }
+
+        if (hideToolbar) {
+            // Remove the status bar from the activity
+            getWindow().setFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN, WindowManager.LayoutParams.FLAG_FULLSCREEN);
+            requestWindowFeature(Window.FEATURE_NO_TITLE);
```
dominicmarmionswrve: The thing about doing it this way is that the title gets removed after the activity is created, thus the titlebar flashes on screen for a split second. The flash can be avoided by configuring the titlebar in the theme rather than programmatically.

If the default is to hide the toolbar, then the default theme on the activity should hide the toolbar, and then programmatically change the theme to one that has a titlebar.

SwrveSDK/src/main/res/values/themes.xml
dominicmarmionswrve, line 1: Think this should be named swrve__themes.xml

Sergio-Mira, line 1: Why? I don't think it collides with a themes.xml provided by the apps

SwrveSDK/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,14 @@
...
+<resources>
+    <style name="Theme.InAppMessageWithToolbar" parent="Theme.InAppMessage">
+        <item name="android:windowNoTitle">false</item>
+    </style>
+
+    <style name="Theme.InAppMessage" parent="android:Theme">
```
dominicmarmionswrve: I think the parent of Theme.InAppMessage needs to be @android:style/Theme.NoTitleBar.Fullscreen and parent of 
Theme.InAppMessageWithToolbar should be @android:style/Theme

And there should be the same styles in a values-v14 folder but with parents of :
@android:style/Theme.Holo.Light
@android:style/Theme.Holo.Light.NoActionBar.Fullscreen

If its done this way then the following line is not needed in SwrveInAppMessageActivity:
https://github.com/Swrve/swrve-android-sdk/pull/195/files#diff-68404ffda1fc09812527ef3d28582b28R83

SwrveSDK/src/main/res/values/themes.xml
dominicmarmionswrve, line 1: Sorry, I thought it did, but I could be wrong. If it doesn't, then disregard! :)

dominicmarmionswrve:
🚀

PR #199, opened 2016-06-08
Add proguard rules to avoid obfuscating of model classes

silin:
If client app uses proguard with agressive rules and obfuscating, all fields in model classes are obfuscated as well. Because of this GSON cannot parse them. That can causes issues in different places, e.g. In-Apps are not working, because `Trigger` entity is not parsed. And other cases

The pull request adds needed proguard rules. But you guys have to verify if the proguard file is added
to .aar file before uploading to public repository (usually it should)

### DIFF
--- a/SwrveSDKCommon/proguard-rules.pro
+++ b/SwrveSDKCommon/proguard-rules.pro
@@ -15,3 +15,11 @@
 #-keepclassmembers class fqcn.of.javascript.interface.for.webview {
 #   public *;
 #}
+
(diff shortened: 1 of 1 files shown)


dominicmarmionswrve:
Thanks @silin for this. We'll look at merging this into our next release in the coming weeks.

Sergio-Mira:
A fix was released with 4.5 in the past weeks. Let us know how it goes!

PR #209, opened 2016-06-29
Open all links from the webview in the browser

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-13024


Sergio-Mira:

dominicmarmionswrve:
🚀

PR #211, opened 2016-06-29
12889 - redraw rounded corners if different radius applied on new page

dominicmarmionswrve:
redraw rounded corners if different radius applied on new page

https://swrvedev.jira.com/browse/SWRVE-12889

@Sergio-Mira please


Sergio-Mira:
🚀

PR #213, opened 2016-07-01
13274 - Add sensor rotation configuration to Conversation SDK

seaders:
https://swrvedev.jira.com/browse/SWRVE-13274
@dominicmarmionswrve & @Sergio-Mira


Sergio-Mira:
🚀

dominicmarmionswrve:
@seaders Do all devices have a sensor?

seaders:
I don't think all do, but at most it'll just ignore it and rotation won't work one way or another.  I think it's more a build system setup thing.  I'll check for the rest, but there's something that's ensuring rotation filters to Activities on native Android, that's not there on Unity Android.

PR #219, opened 2016-08-19
13697 onCreate should be protected

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-13697


dominicmarmionswrve:
🚀 (worth checking we haven't done the same elsewhere)

Sergio-Mira:
Checked and couldn't find any other issues with onCreate.

PR #220, opened 2016-08-19
13631 Disable referrer id by default

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-13631

### DIFF
--- a/SwrveSDK/src/main/AndroidManifest.xml
+++ b/SwrveSDK/src/main/AndroidManifest.xml
@@ -16,14 +16,6 @@
             android:theme="@style/Theme.InAppMessage">
         </activity>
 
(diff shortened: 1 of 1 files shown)


Sergio-Mira:

dominicmarmionswrve:
I think another way to do this work be to use the "enabled=false" flag in the manifest.

```
    <receiver
        android:name="com.swrve.sdk.SwrveInstallReferrerReceiver"
        android:exported="true" 
        android:enabled="false">
```

Then create a new api to programmatically enable the Referrer which customers can then call?

dominicmarmionswrve:
🚀  (So long as we add docs on how to add this back)

PR #224, opened 2016-08-23
[android] Fix android google build error

dominicmarmionswrve:
Fix build error. Removed test testProcessIntentWithDeeplink which is already covered in SwrvePushEngageReceiverTest

@Sergio-Mira please


Sergio-Mira:
🚀

PR #225, opened 2016-08-25
13712 Update dependencies / tools and remove dependency on support v7

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-13712


seaders:
🚀 gogogo!

PR #226, opened 2016-08-26
Demos compile out of the box, core and google versions of demo app

Sergio-Mira:


Sergio-Mira:
Targeting 4.7 here https://github.com/Swrve/swrve-android-sdk/pull/227

PR #228, opened 2016-08-30
Add test attributions

Sergio-Mira:


dominicmarmionswrve:
We only use guava in one test, and it's not really necessary. I could remove it easily and we wouldn't need that dependency.....what you think?

Sergio-Mira:
Yep, would be good if it is not used in many places!

dominicmarmionswrve:
🚀  after https://github.com/Swrve/swrve-android-sdk/pull/232

PR #230, opened 2016-09-06
13863 adm manifest api key

JohnOKane:


build.gradle
```diff
@@ -11,6 +11,7 @@ buildscript {
 allprojects {
     repositories {
         mavenCentral()
+        mavenLocal()
```
Sergio-Mira, line 14: Surprised it is needed when you are pointing to a local jar

SwrveSDK/src/amazon/java/com/swrve/sdk/SwrveEmpty.java
```diff
@@ -18,4 +19,12 @@ protected SwrveEmpty(Context context, String apiKey) {
     //@Override
     //public void setPushNotificationListener(ISwrvePushNotificationListener pushNotificationListener) {
     //}
+
+    @Override
+    public void iapPlay(String productId, double productPrice, String currency, String purchaseData, String dataSignature) {
```
Sergio-Mira, line 24: iapPlay is for Google play, not needed in this flavour. It is probably forcing you because it might be in an interface or similar, remove

JohnOKane: iapPlay is in ISwrveBase - didn't want to push it down to the ISwrve interface in the google flavour in case it affected much other code. Could do?

build.gradle
JohnOKane: To return to before this is finished.

PR #231, opened 2016-09-07
13842 Dialer fix

Sergio-Mira:
https://swrvedev.jira.com/browse/SWRVE-13842


Sergio-Mira:
Merged into 4.6.1 hot fix

PR #236, opened 2016-09-12
[android] updated markdown for samples

adam-govan:
updated the markdown to remove the 404 around the samples page

@seaders / @Sergio-Mira / @dominicmarmionswrve


Sergio-Mira:
🚀

PR #240, opened 2016-09-29
14162_asset_manager_4_6_1

dominicmarmionswrve:
Added a SwrveAssetManager which is the same download asset functionality from SwrveImp, but now easier to test and swap out.

Also added tests to for this for which there was none.

https://swrvedev.jira.com/browse/SWRVE-14162

@Sergio-Mira @seaders @adam-govan

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -1,5 +1,6 @@
 package com.swrve.sdk;
 
+import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.Context;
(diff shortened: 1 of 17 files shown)


SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -118,7 +119,7 @@ protected T init(final Activity activity) throws IllegalArgumentException {
             appStoreURLs = new SparseArray<String>();
 
             // Find cache folder
```
adam-govan: change comment here

adam-govan:
🚀

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
dominicmarmionswrve: Will do!

PR #242, opened 2016-10-11
14106 conv sty apply native sty android

dominicmarmionswrve:
Incremented the conversation version number
Apply styling to buttons and multi input components

https://swrvedev.jira.com/browse/SWRVE-14106

@Sergio-Mira please


SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationButton.java
```diff
@@ -31,6 +35,11 @@ public ConversationButton(Context context, ButtonControl model, int defStyle) {
         initBackgroundColorStates();
         setLines(1);
         setEllipsize(TextUtils.TruncateAt.END);
+        if (conversationVersion >= 4) { // text/font styling added in v4
```
Sergio-Mira: I don't think the button needs to know about the conversation version, just that there are font styles in its style model.

SwrveConversationSDK/src/main/res/values/swrve__themes.xml
```diff
@@ -13,12 +13,12 @@
...
         <item name="conversationContentTrayBottomMargin">@dimen/swrve__content_tray_bottom_margin</item>
         <item name="conversationContentDescriptionStyle">@style/cio__description_item</item>
-        <item name="conversationInputTextBoxDefaultStyle">@style/cio__edit_box</item>
+        <!--<item name="conversationInputTextBoxDefaultStyle">@style/cio__edit_box</item>-->
         <item name="conversationTextContentDefaultStyle">@style/cio__content_textview</item>
         <item name="conversationDialogContentDefaultStyle">@style/cio__dialog_textview</item>
         <item name="conversationDialogButtonDefaultStyle">@style/cio__dialog_button</item>
-        <item name="conversationInputMultiValueItemStyle">@style/cio__mv_item</item>
-        <item name="conversationInputMultiValueDescriptionStyle">@style/cio__mv_description_item</item>
+        <!--<item name="conversationInputMultiValueItemStyle">@style/cio__mv_item</item>-->
```
Sergio-Mira: Remove all these

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
```diff
@@ -942,6 +948,55 @@ protected boolean downloadAssetSynchronously(final String assetPath) {
         return false;
     }
 
+    protected boolean downloadFontAssetSynchronously() {
+
+        String cdnRoot = "http://content-cdn.swrve.com/messaging/fonts/";
```
Sergio-Mira: Tests should not contact the net. Copy the font file into the test project.

SwrveConversationSDK/src/main/java/com/swrve/sdk/SwrveConversationConstants.java
```diff
@@ -0,0 +1,9 @@
+package com.swrve.sdk;
+
+/**
+ * Used internally
+ */
+public class SwrveConversationConstants {
+    public static final String EXTRA_CONVERSATION_KEY = "conversation";
+    public static final String EXTRA_CONVERSATION_VERSION_KEY = "conversation_version";
```
Sergio-Mira: Is this used for anything? Can only see writes to this key.

SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationButton.java
dominicmarmionswrve: Yeah, I thought about this, and this is a SwrveConversationButton, that already takes a ButtonControl object so passing in a conversationversion isn't such a big deal here IMO.

It needs to be able to render old style of conversations and new ones, hence passing the version in was needed. 

This is WIP, as I'm removing the default styles and applying them programmatically so all configuration is in one place. More to come.....

SwrveConversationSDK/src/main/res/values/swrve__themes.xml
dominicmarmionswrve: Yup, WIP....

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
dominicmarmionswrve: This is the default cdnroot which was already there, so I kept it in there. But I agree, and would prefer to remove  the default. Will do!!

SwrveConversationSDK/src/main/java/com/swrve/sdk/SwrveConversationConstants.java
dominicmarmionswrve: Its used as a key to put SwrveConversation serialized object into an extra in the intent to start conversationactivity. Used by unity also. So best to use a constant.

dominicmarmionswrve: oops, you talking about the EXTRA_CONVERSATION_VERSION_KEY, not the EXTRA_CONVERSATION_KEY! :)
But I think its the same point.

SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationButton.java
Sergio-Mira: Style should have values for those new properties or null / not set. The button should not know about the conversation version that caused those style changes in my opinion.

dominicmarmionswrve: Fair point.......will work on that.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveImp.java
dominicmarmionswrve: Will remove this as part of the download font jira.

SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationButton.java
dominicmarmionswrve: changes made

SwrveConversationSDK/src/main/java/com/swrve/sdk/SwrveConversationConstants.java
dominicmarmionswrve: Yup, these are not required as public constants. Moved to ConversationActivity.

SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/ui/ConversationFragment.java
```diff
@@ -221,6 +223,15 @@ private void renderControls(Activity activity) {
         }
     }
 
+    private void applyButtonStyles(ConversationStyle style) {
+        Typeface typeface = getTypeface(style, SwrveConversationConstants.DEFAULT_BUTTON_TYPEFACE);
+        style.setTypeface(typeface);
+
+        if(conversationVersion < 4) {
```
Sergio-Mira: We should still render as before, change the text size of the button based on the text_size attribute if present

dominicmarmionswrve: Nope. version 4+ will always have the text size. We should always use that.

Only under version 4's need a default.

PR #244, opened 2016-10-14
14097 font download android

dominicmarmionswrve:
Added font downloading and using the new cdnPaths.
Changed the Type on Atoms from a String to an ENUM.

https://swrvedev.jira.com/browse/SWRVE-14097

### DIFF
--- a/SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/engine/model/ConversationAtom.java
+++ b/SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/engine/model/ConversationAtom.java
@@ -1,23 +1,28 @@
 package com.swrve.sdk.conversations.engine.model;
 
 
+import com.google.gson.annotations.SerializedName;
 import com.swrve.sdk.conversations.engine.model.styles.ConversationStyle;
 
 import java.io.Serializable;
 
 public abstract class ConversationAtom implements Serializable {
-    public static final String TYPE_CONTENT_HTML = "html-fragment";
-    public static final String TYPE_CONTENT_IMAGE = "image";
-    public static final String TYPE_CONTENT_VIDEO = "video";
-    public static final String TYPE_CONTENT_SPACER = "spacer";
-    public static final String TYPE_INPUT_MULTIVALUE = "multi-value-input";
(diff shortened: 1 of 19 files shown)


dominicmarmionswrve:
@seaders @Sergio-Mira @adam-govan please

SwrveConversationSDK/src/main/java/com/swrve/sdk/conversations/engine/model/ConversationAtom.java
```diff
@@ -1,23 +1,28 @@
...
 public abstract class ConversationAtom implements Serializable {
-    public static final String TYPE_CONTENT_HTML = "html-fragment";
-    public static final String TYPE_CONTENT_IMAGE = "image";
-    public static final String TYPE_CONTENT_VIDEO = "video";
-    public static final String TYPE_CONTENT_SPACER = "spacer";
-    public static final String TYPE_INPUT_MULTIVALUE = "multi-value-input";
-    public static final String TYPE_INPUT_STARRATING = "star-rating";
+
+    public enum TYPE {
+        @SerializedName("html-fragment")        CONTENT_HTML,
```
Sergio-Mira, line 12: 👍

SwrveSDK/src/main/java/com/swrve/sdk/conversations/SwrveConversation.java
```diff
@@ -60,6 +92,20 @@ public boolean areAssetsReady(Set<String> assetsOnDisk) {
         return true;
     }
 
+    private boolean isAssetInCache(Set<String> assetsOnDisk, String asset) {
```
Sergio-Mira: ~ isImageAsset?

Sergio-Mira:
🚀 (does it do any SHA1 check? note that if we do I've updated the upload to be companyId-SHA1.extension)

dominicmarmionswrve:
The sha is checked here https://github.com/Swrve/swrve-android-sdk/blob/14097_font_download_android/SwrveSDKCommon/src/main/java/com/swrve/sdk/SwrveAssetsManagerImp.java#L123

But it assumes the whole value is encrypted, including the companyid and extension, which would be inconsistent with server. Lets chat about this.....

PR #247, opened 2016-11-03
14545 [android] type based api for dates

adam-govan:
Provided a new public api for customers to add their own user updates with Date objects. 

This date gets converted into a UTC ISO8601 format and passed as a string for backend. 

@Sergio-Mira / @seaders / @dominicmarmionswrve can you take a look please?

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
@@ -40,7 +40,7 @@ public interface ISwrveBase<T, C extends SwrveConfigBase> {
 
     void userUpdate(Map<String, String> attributes);
 
-    void userUpdateDate(String name, Date date);
+    void userUpdate(String name, Date date);
 
     void iap(int quantity, String productId, double productPrice, String currency);
 
--- a/SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
@@ -40,6 +40,8 @@ public interface ISwrveBase<T, C extends SwrveConfigBase> {
 
     void userUpdate(Map<String, String> attributes);
 
+    void userUpdateDate(String name, Date date);
+
     void iap(int quantity, String productId, double productPrice, String currency);
 
     void iap(int quantity, String productId, double productPrice, String currency, SwrveIAPRewards rewards);
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -45,6 +45,8 @@ import org.json.JSONObject;
 import java.io.File;
 import java.io.UnsupportedEncodingException;
 import java.net.HttpURLConnection;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Collections;
@@ -55,6 +57,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 
 import static com.swrve.sdk.SwrveCampaignDisplayer.DisplayResult.CAMPAIGN_WRONG_ORIENTATION;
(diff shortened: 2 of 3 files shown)


SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -317,6 +320,41 @@ protected void _userUpdate(Map<String, String> attributes) {
         queueEvent("user", parameters, null);
     }
 
+    protected void _userUpdateDate (Map<String, Date> attributes) {
```
Sergio-Mira:
~~~ if we are only going to have one key we could remove the for and move this logic into the other _userUpdateDate

Sergio-Mira:
🚀

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -317,6 +320,41 @@ protected void _userUpdate(Map<String, String> attributes) {
...
+        for(String key : attributes.keySet()) {
+            Date object = attributes.get(key);
+            String resultantDate = getStringFromDate(object);
+            resultAttributes.put(key, resultantDate);
+        }
+
+        try {
+            parameters.put("attributes", new JSONObject(resultAttributes));
+        }catch (NullPointerException ex) {
+            SwrveLogger.e(LOG_TAG, "JSONException when encoding user attributes", ex);
```
Sergio-Mira, line 333: This should stop the method, not continue with queueEvent

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
adam-govan: I was thinking if we package the end result into a map of attributes. might as well have the option to iterate it. Doesn't have to be public facing I think it'll be useful if we want to batch these down the line.

SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
```diff
@@ -40,6 +40,8 @@
 
     void userUpdate(Map<String, String> attributes);
 
+    void userUpdateDate(String name, Date date);
```
dominicmarmionswrve: Now that the other api is gone, it would be nicer if it was just userUpdate(String name, Date date) ?
(ie change name from userUpdateDate to userUpdate)

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -317,6 +320,41 @@ protected void _userUpdate(Map<String, String> attributes) {
         queueEvent("user", parameters, null);
     }
 
+    protected void _userUpdateDate (Map<String, Date> attributes) {
+        Map<String, Object> parameters = new HashMap<String, Object>();
+        Map<String, String> resultAttributes = new HashMap<String, String>();
+
+        for(String key : attributes.keySet()) {
```
dominicmarmionswrve: Why iterate through a map when only a string and date are going to be used?

SwrveSDK/src/main/java/com/swrve/sdk/ISwrveBase.java
adam-govan: @dominicmarmionswrve  that's not a bad idea :) I'll do that now.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
dominicmarmionswrve: This overcomplicates something which is should be a straight forward IMO.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
adam-govan: added `return;` to the catch which should exit it out

dominicmarmionswrve: Better to check for null rather than catching a NullPointerException.

SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -317,6 +320,32 @@ protected void _userUpdate(Map<String, String> attributes) {
...
+            return;
+        }
+
+        queueEvent("user", parameters, null);
+    }
+
+    private String getStringFromDate(Date date) {
+
+        TimeZone timezone = TimeZone.getTimeZone("UTC");
+        DateFormat dateformat = new SimpleDateFormat("yyyy-dd-MM'T'HH:mm:ss.SSS'Z'");
```
dominicmarmionswrve, line 343: Might be better to initialize this as part of init of sdk? Creating new SimpleDateFormat can be expensive, I think.

PR #248, opened 2016-11-03
14628 - using single lock for event storage read/write

dominicmarmionswrve:
WIP  - tests and other locking bits need investigation
WIP - might be merged into 4_6_2.....to be decided yet....


jmason:
I don't know how viable it'd be, but it may be worth annotating the codebase with @Threadsafe, @Immutable, @NonThreadsafe, @GuardedBy annotations, and checking the codebase with Error Prone, which will detect a good number of thread-safety bugs -- we use it a long in the server side code and it's a godsend.   Also Intelli/J's "Inspect" mode does a great job of this too....

dominicmarmionswrve:
Will that increase the library size? or are these only  used at compile time?

dominicmarmionswrve:
retargeted here:
https://github.com/Swrve/swrve-android-sdk/pull/249

jmason:
'Will that increase the library size? or are these only used at compile time?' -- @dominicmarmionswrve only used at compile time.

PR #249, opened 2016-11-07
14628 - using single lock for event storage read/write

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-14628
(Originally started here https://github.com/Swrve/swrve-android-sdk/pull/248)

Using one lock for sending events MemoryCachedLocalStorage.EVENT_LOCK.
The old SwrveEventsManager had its own lock which was causing dupe events when a flush was done whilst sending events at the same time.

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/localstorage/MemoryCachedLocalStorage.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/localstorage/MemoryCachedLocalStorage.java
@@ -18,7 +18,8 @@ public class MemoryCachedLocalStorage implements IMemoryLocalStorage {
     private ILocalStorage cache;
     private ILocalStorage secondaryStorage;
 
-    private Object eventLock = new Object();
+    public static final Object EVENT_LOCK = new Object();
+
     private Object cacheLock = new Object();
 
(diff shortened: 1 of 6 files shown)


Sergio-Mira:
🚀

dominicmarmionswrve:
@Sergio-Mira @adam-govan can you take a look please?

SwrveSDK/src/main/java/com/swrve/sdk/localstorage/MemoryCachedLocalStorage.java
```diff
@@ -18,7 +18,8 @@
     private ILocalStorage cache;
     private ILocalStorage secondaryStorage;
 
-    private Object eventLock = new Object();
+    public static final Object EVENT_LOCK = new Object();
```
Sergio-Mira, line 21: Could this be passed by the Swrve object instead of being static?

Sergio-Mira:
🚀

SwrveSDK/src/main/java/com/swrve/sdk/localstorage/MemoryCachedLocalStorage.java
dominicmarmionswrve: I guess it could be, but tbh none of this is very elegant. Would like to change a few more things like that.

PR #250, opened 2016-11-08
14646_device_info_sent_repeatedly

dominicmarmionswrve:
https://swrvedev.jira.com/browse/SWRVE-14646

Device info is queued at session start and when reg id changes, instead of every onResume

@Sergio-Mira small one please


Sergio-Mira:
🚀

PR #259, opened 2016-11-30
Use dips as text size units for native atoms

dominicmarmionswrve:
Text size units for native atoms need to be density independent pixels (dip) so that they are the same as webview text sizes which are pixels. This gives a consistent text size across atoms.

@Sergio-Mira please


Sergio-Mira:
🚀

Issue #280, opened 2018-12-17
[Crash] when get a new push notification on the build type which enabled ProGuard


patrickyin:
### Issue
The app build which enabled ProGuard is crashed when a new Swrve push notification comes.

### Cause
After obfuscation, the GSON model class's name or properties' name has been changed, but GSON uses property reflection to map the JSON object to model class. So when the new push comes, the GSON couldn't deserialize the response.

### Solution
Avoid the push notification related models are obfuscated by adding an additional ProGuard rule to the SDK

dominicmarmionswrve:
@patrickyin We'll add a change to address this in our 6.1 release which should be public in next few weeks. Thanks for logging the issue.

dominicmarmionswrve:
@patrickyin a fix for this issue is now part of release 6.1 

I'll close this issue in a couple of days if you've no further comments. Thanks again for reporting it!

patrickyin:
Thanks @dominicmarmionswrve

PR #281, opened 2018-12-17
Fix push notification crash by updating proguard-rules

patrickyin:
This PR fixes the issue #280

### DIFF
--- a/SwrveSDK/proguard-rules.pro
+++ b/SwrveSDK/proguard-rules.pro
@@ -3,3 +3,4 @@
 -keepattributes *Annotation*
 -dontwarn sun.misc.**
 -keep class com.swrve.sdk.messaging.model.** { *; }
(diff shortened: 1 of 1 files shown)


Sergio-Mira:
Thanks for sharing this with us @patrickyin. We have Pro Guard enabled in our internal app and we are not seeing this problem, could you maybe share with us how you have the pro guard rules setup for this to happen?
We'll add this pro guard rule to the next release.

patrickyin:
@Sergio-Mira Thanks for your reply. The model class in Swrve SDK causes our app crash when a new Swrve push notification in(The GSON in Swrve SDK couldn't deserialize the payload of the push notification), once I added this line to the app's ProGuard rule file, the bug has been fixed. But if you can cover it in SDK rather than letting every app to skip obfuscate the Swrve model class, that would be awesome.

Sergio-Mira:
This has been released as part of 7.3.2

Issue #285, opened 2019-04-10
Multiple StrictMode warnings during app initialization


gmerinojimenez:
I'm testing the SDK integration in our app and I'm noticing a lot of strict mode warnings during the startup of the application (11 in my latest tests).

Checking the code, looks like that during the init() method, executed in the onActivityCreated callback method, which runs on the UI thread, there are multiple accesses to disk (initialize the cache directory, open and read from the database...). Are you aware of this? According to the warnings the performance penalty seems noticeable. 

My suggestion is, as google says, move the IO work to a background job.

A few examples:
`04-10 19:10:35.697 19621 19621 D StrictMode: StrictMode policy violation; ~duration=236 ms: android.os.strictmode.DiskReadViolation
04-10 19:10:35.697 19621 19621 D StrictMode: 	at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1504)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:251)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at java.io.File.exists(File.java:1168)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at android.app.ContextImpl.getDataDir(ContextImpl.java:2301)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at android.app.ContextImpl.getCacheDir(ContextImpl.java:695)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at android.content.ContextWrapper.getCacheDir(ContextWrapper.java:269)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at com.swrve.sdk.SwrveBase.initCacheFolder(SwrveBase.java:240)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at com.swrve.sdk.SwrveBase.init(SwrveBase.java:131)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at com.swrve.sdk.SwrveBase.onCreate(SwrveBase.java:103)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at com.swrve.sdk.SwrveBase.onActivityCreated(SwrveBase.java:1848)
04-10 19:10:35.697 19621 19621 D StrictMode: 	at android.app.Application.dispatchActivityCreated(Application.java:248)
[...]

PR #292, opened 2020-03-17
Fixing StrictMode

gmerinojimenez:
In this PR I'm fixing [the StrictMode warnings](https://github.com/Swrve/swrve-android-sdk/issues/285) produced during Swrve initialization.

The idea is to move initialization process to a background thread. As suggested [here](https://developer.android.com/topic/performance/threads). Synchrony should be ensured since the init method is already synchronized.

### DIFF
--- a/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
+++ b/SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
@@ -94,18 +94,28 @@ public abstract class SwrveBase<T, C extends SwrveConfigBase> extends SwrveImp<T
         return null;
     }
 
-    protected T onCreate(final Activity activity) throws IllegalArgumentException {
+    protected void onCreate(final Activity activity) throws IllegalArgumentException {
         if (destroyed) {
             destroyed = initialised = false;
             bindCounter.set(0);
         }
         if (!initialised) {
             // First time it is initialized
-            return init(activity);
+            initInBackground(activity);
+            return;
         }
 
         bindToContext(activity);
-        return (T) this;
+    }
+
+    private void initInBackground(Activity activity) {
+        Thread backgroundThread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                init(activity);
+            }
+        });
+        backgroundThread.start();
     }
 
     protected synchronized T init(final Activity activity) throws IllegalArgumentException {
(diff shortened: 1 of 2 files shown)


SwrveSDK/src/main/java/com/swrve/sdk/SwrveBase.java
```diff
@@ -94,18 +94,28 @@ public static String getVersion() {
...
-    protected T onCreate(final Activity activity) throws IllegalArgumentException {
+    protected void onCreate(final Activity activity) throws IllegalArgumentException {
         if (destroyed) {
             destroyed = initialised = false;
             bindCounter.set(0);
         }
         if (!initialised) {
             // First time it is initialized
-            return init(activity);
+            initInBackground(activity);
```
Dimezis, line 104: You can not do it just like that. What if user tries to access any Swrve methods before it gets fully initialized in the background?

gmerinojimenez: Actually that can happen also without this change. Swrve is initialized with the first activity creation, but an app can be started without user interaction (receiving a push, for example), so Swrve code can be executed also without calling init().

My point is: If that is a problem, it should be prevented outside of the scope of this review.

Sergio-Mira: The SDK is meant to be used from user activity, and after the first Activity has been created.

Unfortunately many things will have to move to be async when we move the initialisation to be in a separate thread.

Dimezis: > Actually that can happen also without this change. Swrve is initialized with the first activity creation, but an app can be started without user interaction (receiving a push, for example), so Swrve code can be executed also without calling init().
> 
> My point is: If that is a problem, it should be prevented outside of the scope of this review.

Before this change, you could safely assume that calling any Swrve methods in any Activity callback is safe. Now it's not. 
I don't think it's a way to go, even though I'm really waiting for this change

gmerinojimenez: In an activity callback, but for example calling a swrve method when a swrve push arrives will be as unsafe as you suggest.

@Sergio-Mira, there is no reference in the documentation saying that

Sergio-Mira: Will talk to the team to make it more explicit in the docs

